Issue #369, opened 2015-08-03
Multiply or Divide transform of Position or Length is incorrect


ericfont:
If I try to perform a Multiply or Divide transform of position or length, the midi events are incorrectly transformed.  When looking at the code in MidiTransformerDialog::transformEvent, I notice the corresponding switch statements are incorrectly using "var" as input to the these transformations, but "pos" or "len" should be used instead.  I have fixed this, compiled and tested, and verified that it is now working correctly, so please accept my pull request.

ericfont:
pull request: https://github.com/muse-sequencer/muse/pull/370

PR #370, opened 2015-08-03
fix issue #369: mul/div transform len or pos

ericfont:
If I try to perform a Multiply or Divide transform of position or length, the midi events are incorrectly transformed.  When looking at the code in MidiTransformerDialog::transformEvent, I notice the corresponding switch statements are incorrectly using "var" as input to the these transformations, but "pos" or "len" should be used instead.  I have fixed this, compiled and tested, and verified that it is now working correctly, so please accept my pull request.


danvd:
Hi Eric,

Thanks for this patch, I'll merge it now.

Regards,
Andrew

PR #442, opened 2015-12-28
fix build

derselbst:
missing forward decl may break build


danvd:
Hi Tom,

Though full declaration of frient class inside another class can be considered as forward declaration, I've read that some version of gcc may fail if no previous forward declaration was found.

Have you got such errors or it's just in case?

Anyway, patch will be merged now.

derselbst:
in my case gcc 4.8.3 failed to compile it

danvd:
OK, thanks for the info!

PR #443, opened 2015-12-28
fix build if ENABLE_DSSI==FALSE

derselbst:
if built without DSSI support, there wont be a member DssiSynth in MusECore


danvd:
Thanks, Tom!
Just merged.

PR #447, opened 2015-12-29
fix build without having liblo installed

derselbst:
liblo is still optional, right? So just include stdint.h if there is no osc support, as it would be included by lo_osc_types.h anyway.


danvd:
ok :)

PR #454, opened 2016-01-08
Add Korg Wavestation EX instrument definition

songo:
Patches (performances), drummaps and controller definitions for Korg Wavestation EX synthesizer (also applicable to non-EX Wavestation with exception of drummaps and  factory bank 3). Tested on a real hardware.


danvd:
Ok. thanks - merged :)

Issue #458, opened 2016-01-17
Empty MTrk chunk breaks MIDI Import


derselbst:
When importing a MIDI File, which has a MTrk chunk of zero size, some events in the follwing MTrk chunk might be ommitted, resulting in uncorrect delta-delays. See an example MIDI here: 

[https://drive.google.com/open?id=0B7HVOcdsHTi3QU8tS092ZlU4eVU](https://drive.google.com/open?id=0B7HVOcdsHTi3QU8tS092ZlU4eVU)

When importing the MIDI you should get the follwing result:
![musemidiwrong](https://cloud.githubusercontent.com/assets/8152480/12376855/b57a8ee4-bd08-11e5-958a-008ab6d884db.png)

Note the MThd: in order to import this file, you have to set number-of-trks to only 2, although the file contains obviously 3 MTrks.

Removing the empty MTrk chunk at offset 0x388 should cause correct import result:
![musemidicorrect](https://cloud.githubusercontent.com/assets/8152480/12376879/7650efb4-bd09-11e5-9a09-481a2d5ef360.png)

derselbst:
One minor question by the way: Is it normal the way you nowadays display ctrl-events like mainvolume, pitch or pan in those track canvases (or what do you call them?) - is this a bug or feature?
Back in Muse 2.1 they were displayed like:
![musetrackcanvas](https://cloud.githubusercontent.com/assets/8152480/12376935/bd18b416-bd0b-11e5-9a54-3b9de5d875c3.png)

terminator356:
On January 17, 2016 02:21:12 AM Tom M. wrote:
> One minor question by the way: Is it normal the way you nowadays display
> ctrl-events like mainvolume, pitch or pan in those track canvases (or what
> do you call them?) - is this a bug or feature? Back in Muse 2.1 they were
> displayed like:
> ![musetrackcanvas](https://cloud.githubusercontent.com/assets/8152480/123769
> 35/bd18b416-bd0b-11e5-9a54-3b9de5d875c3.png)
> 
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/muse-sequencer/muse/issues/458#issuecomment-172309970

Hi, I've downloaded the file and will take a look.

Thanks.
Tim.

terminator356:
On January 20, 2016 03:56:12 AM you wrote:
> On January 17, 2016 02:21:12 AM Tom M. wrote:
> > One minor question by the way: Is it normal the way you nowadays display
[...]

PR #461, opened 2016-01-22
another midi import fix, addresses #458

derselbst:
...and might even fix it


derselbst:
fixed in aeaecfcf9f4ae708899ac237d00a22f8728ca28f and 6b0c76e38530f5ebca9d8168946956e55656a4f1

PR #464, opened 2016-01-25
Polish translation file: a major update

songo:
Hi!
This is a major update of the Polish translation file. Unlike a previous one, it's almost complete translation for `3.0` (some plugins/plugin parameters  have been omitted because I doubt whether these need translation at all). Perhaps it  needs some polishing or minor changes, but it's ready for use IMO.

Long live MusE :)
       k.


spamatica:
Many thanks for the update and long live MusE! :)

PR #559, opened 2017-01-25
Replace deprecated WFlags references with WindowFlags

RomanHargrave:
As of this pr, muse doesn't compile with the latest Qt version due to the deprecated `Qt::WFlags` type having been removed.

It was synonymous to `Qt::WindowFlags`. I replaced the few instances in muse, and the build was able to continue.

### DIFF
--- a/muse3/muse/instruments/midictrledit.cpp
+++ b/muse3/muse/instruments/midictrledit.cpp
@@ -102,7 +102,7 @@ void MidiControllerEditDialog::addControllerToView(MidiController* mctrl)
 //   MidiControllerEditDialog
 //---------------------------------------------------------
 
-MidiControllerEditDialog::MidiControllerEditDialog(QWidget* parent, const char* name, bool modal, Qt::WFlags fl)
+MidiControllerEditDialog::MidiControllerEditDialog(QWidget* parent, const char* name, bool modal, Qt::WindowFlags fl)
    : MidiControllerEditDialogBase(parent, name, modal, fl)
       {
       _lastPort = midiPortsList->currentItem();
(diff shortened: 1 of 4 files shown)


spamatica:
Awesome, merged.

Thanks

terminator356:
On Wednesday, January 25, 2017 12:42:48 AM EST Roman Hargrave wrote:
> As of this pr, muse doesn't compile with the latest Qt version due to the
> deprecated `Qt::WFlags` type having been removed.
> 
> It was synonymous to `Qt::WindowFlags`. I replaced the few instances in
> muse, and the build was able to continue. You can view, comment on, or
> merge this pull request online at:
> 
>   https://github.com/muse-sequencer/muse/pull/559
> 
> -- Commit Summary --
> 
>   * Replace deprecated WFlags references with WindowFlags, now compiles on
> latest Debian Testing/Stretch
> 
> -- File Changes --
> 
>     M muse3/muse/instruments/midictrledit.cpp (2)
>     M muse3/muse/instruments/midictrledit.h (2)
>     M muse3/muse/mplugins/rhythm.cpp (4)
>     M muse3/muse/mplugins/rhythm.h (2)
> 
> -- Patch Links --
> 
> https://github.com/muse-sequencer/muse/pull/559.patch
> https://github.com/muse-sequencer/muse/pull/559.diff

Thanks!
Our age is showing, eh?

Not sure how these two modules got included in the build.
Probably an oversight.
They should be removed from cmake build, but kept for historical reasons.

midictrledit.cpp contains some old code we don't use ATM.
It is supposed to allow new user-definable controllers to be added.
Some day I hope (need actually) to restore it.

rhythm.cpp is also has old code we don't use ATM.
It is an interesting module that was supposed to allow
 automatically generated rhythm.
But, well it is so old we probably will have a tough time
 ever reactivating it.
Still, I like to keep it around for a rainy day.

Tim.

PR #600, opened 2018-01-05
Midi transform position randomize

oswald2:
Hi,

As discussed on the mailing list I added the functionality for randomizing the MIDI note position. The Transfrom Dialog got an additional entry in the Position Combo Box which is then subsequently handled in the MidiTransformerDialog code. The change is quite minor.

lg,
Michael


spamatica:
Thanks, merged :)

PR #607, opened 2018-01-19
Midi plugins

oswald2:
Hello,

I did some fixes on the RandomPosition1 MIDI plugin which was in the source tree but was not installed. It now works.
I also added a new plugin RandomizeVelocityRelative, which does the same to the velocity. So instead of using absolute values for a random range like in the Transformer Dialog (like e.g. "generate velocities between 90 and 110), this plugin takes the current velocities of notes and just jitters them with a random value where the maximum spread can be adjusted. Basically it is the RandomizePosition1 plugin changed from the tick to the velocity.

I find them quite useful.

lg,
Michael


spamatica:
Wow thanks, adding :)

PR #614, opened 2018-08-03
Add idf for CrocellKit from Drumgizmo

windowsrefund:
Trying to help make [this tutorial](https://www.youtube.com/watch?v=hXDCTmcBmLg) easier to work with since it's not obvious that the mapping has to be created. Also, it would be really nice if the Instrument Editor showed MIDI Note Numbers (0-127) in one of the columns rather than just showing 2 columns of musical notes. It would make future contributions like this easier since, in the case of Drumgizmo, they provide this information in their .xml files.

### DIFF
--- a/muse3/share/instruments/Drumgizmo - CrocellKit.idf
+++ b/muse3/share/instruments/Drumgizmo - CrocellKit.idf
@@ -1,0 +1,149 @@
+<?xml version="1.0"?>
+<muse version="2.1">
+  <MidiInstrument name="Drumgizmo CrocellKit">
+    <PatchGroup name="Drums">
+      <Patch name="Drums" drum="1" />
+    </PatchGroup>
+    <Drummaps>
+      <entry>
+        <drummap>
+          <entry pitch="35">
+            <name>Kick L</name>
+          </entry>
+          <entry pitch="36">
+            <name>Kick R</name>
+          </entry>
+          <entry pitch="37">
+            <name>Snare Rim</name>
+          </entry>
+          <entry pitch="38">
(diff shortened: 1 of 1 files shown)


windowsrefund:
Given the comments in the thread below, I now realize this is a dumb PR as .idf files are not appropriate for synths?

https://linuxmusicians.com/viewtopic.php?p=95407#p95407

spamatica:
Sorry, a bit late to the game. 
Yes, idf files should not be necessary with soft synths as they should already expose all relevant parameters to the host (MusE). 
What exactly were you trying to fix with the idf?
There's also the concept of drumlists/drummaps (right click on the name of a drumtrack in the arranger to select one) where naming and mapping of specific drums can be defined.

terminator356:
Wait! Your work is not entirely in vain.
Some plugins can be run stand-alone from a command line,
 and you may wish to connect MusE to them.
Other plugins might be loaded in an entirely different application, say Ardour,
 which you may wish to connect MusE to.

It is in these situations where an .idf file DOES help.
Because now MusE has absolutely no knowledge of what you are connecting to,
 and does not have the luxury of obtaining instrument information from the plugin.
MusE is now just blindly sending midi to the destination, much like sending to
 a real external hardware midi device.
So here an .idf file is essential to define all the characteristics of the external 
 application, plugin or device. All the patch names, controllers etc. need to be defined.

Tim.

cyberic99:
These idf files look a bit like Ardour's patchfiles: https://github.com/Ardour/ardour/tree/master/patchfiles

Is there a way to use a common format?

PR #615, opened 2018-08-04
Replace idf with map

windowsrefund:
A usable map for CrocellKit from Drumgizmo

### DIFF
--- a/muse3/share/instruments/Drumgizmo - CrocellKit.idf
+++ b/muse3/share/instruments/Drumgizmo - CrocellKit.idf
@@ -1,0 +1,149 @@
+<?xml version="1.0"?>
+<muse version="2.1">
+  <MidiInstrument name="Drumgizmo CrocellKit">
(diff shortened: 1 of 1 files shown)


windowsrefund:
Hmm... something I just noticed about these drum maps. The order is not maintained in the schema. I know this now because I just reorganized it (HiHats together, Toms together, etc) and saved a fresh copy. My intention was to add the improved map to this PR. No change... I wonder if I'll see that MusE itself maintains the order in the Drum Editor? Interesting.

terminator356:
Shall I merge this or wait a bit?
As mentioned, it's OK if you want to leave the .idf file in there,
 they are actually needed if driving the stand-alone version of DrumGizmo
 from MusE. I'd actually recommend leaving it in there in this particular case
 since some plugins don't have stand-alone versions, but DG happens to have one.

No harm in doing so. Tough work was done, might as well save it?

windowsrefund:
done

PR #624, opened 2018-09-04
Midi note map

windowsrefund:
At this time, MusE does not make it very easy to see the midi note numbers in a drum map. Therefore, I decided to work around the issue by creating a map showing the numbers where instrument names are usually displayed. Having the ability to quickly change to a map like this makes things easier when it comes to transitioning/auditing different drumkits where instruments are mapped using odd numbers.


terminator356:
Hm, was never merged, sorry about that. Done.

PR #630, opened 2018-10-17
fluidsynth: Follow API change introduced in fluidsynth 2.0.0

schnitzeltony:
Build tested with fluidsynth 1.1.11 and 2.0.1

### DIFF
--- a/muse3/synti/fluidsynth/fluidsynti.cpp
+++ b/muse3/synti/fluidsynth/fluidsynti.cpp
@@ -155,16 +155,22 @@ FluidSynth::~FluidSynth()
           std::cerr << DEBUG_ARGS << "Error unloading soundfont!" << fluid_synth_error(fluidsynth) << std::endl;
       }
         
+#if FLUIDSYNTH_VERSION_MAJOR < 2
       int err = delete_fluid_synth (fluidsynth);
+#else
+      delete_fluid_synth (fluidsynth);
+#endif
       if(gui)
         delete gui;
 
       if (initBuffer)
             delete [] initBuffer;
+#if FLUIDSYNTH_VERSION_MAJOR < 2
       if (err == -1) {
             std::cerr << DEBUG_ARGS << "error while destroying synth: " << fluid_synth_error(fluidsynth) << std::endl;
             return;
-            }            
+            }
+#endif
       }
 
 bool FluidSynth::init(const char* name)
@@ -1368,7 +1374,11 @@ const char* FluidSynth::getPatchName(int i, int, bool /*drum*/) const
       else {
             fluid_preset_t *preset = fluid_synth_get_channel_preset(fluidsynth, i);
             if (!preset) return "<unknown>";
+#if FLUIDSYNTH_VERSION_MAJOR < 2
             return preset->get_name(preset);
+#else
+            return fluid_preset_get_name(preset);
+#endif
             }
       }
 //---------------------------------------------------------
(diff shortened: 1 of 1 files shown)


terminator356:
Hi, thank you. The changes look fine.
I had a different idea but there are problems with it.
Please see the proposed changes in this branch:
https://github.com/muse-sequencer/muse/tree/fluidsynth_v2

I don't like the idea of having to re-configure and build if
 a dependency is up/down-graded.
I'm more in favour of a no-recompile approach whenever possible.
That approach was used successfully with our Jack-1/2 support:
The user can install more than one version of Jack and immediately whichever one the
 ldconfig system sees first is the one MusE uses - _without_ re-compiling MusE.
That was no easy task, as not only are Jack's APIs different but the fundamental
 operations such as timing as well. Our jack.cpp/h module shows how it was done.
It required good old fashioned dynamic symbol look-ups, the only possible solution.

However, when I tried that approach with fluidsynth-2.x I got trouble.

The runtime dynamic linker will refuse to link to a library with a
 DIFFERENT major .so version, IF the library has such .so versions at all.

Thus you can upgrade to a library with a different MINOR version number,
 and the app will still run, but not one with a different MAJOR version number.
 
The following printout shows why my approach failed with fluidsynth-2:
Notice that both Jack versions 0.x and 1.x are installed BUT they have no
 major .so version numbers. So MusE will happily choose whichever one happens
 to be installed and because of our support, will run fine WITHOUT recompilation.
Also, notice that libsndfile also has a local version installed and MusE uses it
 wihtout recompilation.
But notice that fluidsynth DOES have major .so version numbers.
So the runtime linker appears to always choose the one matching the major number
 requested by the executable - notice that in this case MusE compiled wanting
 version libfluidsynth.so.2 simply because I happened to have version 2
 installed at the time. So MusE will ignore the local fluidsynth version 1.x.
 
"sudo ldconfig -p"
libjack.so.0 (libc6,x86-64) => /usr/local/lib64/libjack.so.0
libjack.so.0 (libc6,x86-64) => /usr/lib64/libjack.so.0
libjack.so.0 (libc6) => /usr/lib/libjack.so.0
libjack.so (libc6,x86-64) => /usr/local/lib64/libjack.so
libjack.so (libc6,x86-64) => /usr/lib64/libjack.so

libsndfile.so.1 (libc6,x86-64) => /usr/local/lib64/libsndfile.so.1
libsndfile.so.1 (libc6,x86-64) => /usr/lib64/libsndfile.so.1
libsndfile.so.1 (libc6) => /usr/lib/libsndfile.so.1
libsndfile.so (libc6,x86-64) => /usr/local/lib64/libsndfile.so
libsndfile.so (libc6,x86-64) => /usr/lib64/libsndfile.so

libfluidsynth.so.2 (libc6,x86-64) => /usr/lib64/libfluidsynth.so.2
libfluidsynth.so.1 (libc6,x86-64) => /usr/local/lib64/libfluidsynth.so.1
libfluidsynth.so (libc6,x86-64) => /usr/local/lib64/libfluidsynth.so
libfluidsynth.so (libc6,x86-64) => /usr/lib64/libfluidsynth.so

"ldd /muse-install-prefix/bin/muse3"
libjack.so.0 => /usr/local/lib64/libjack.so.0 (0x00007fdfb6f03000)

"ldd /muse-install-prefix/lib/muse-3.0/synthi/fluid_synth.so"
libfluidsynth.so.2 => /usr/lib64/libfluidsynth.so.2 (0x00007f77034f6000)
libsndfile.so.1 => /usr/local/lib64/libsndfile.so.1 (0x00007f77024cd000)

Therefore, based on that, my approach is a big fail here.
Therefore, since we cannot do anything about it at runtime,
I will merge your patch and we'll be done with it.
It's just too bad that someone who builds MusE locally will have to
 recompile as soon as fluidsynth version 2.x comes down their distro's
 package pipeline.

Cheers.
Tim.

PR #634, opened 2019-01-19
Run properly on HiDPI screens

hfiguiere:
Without this the UI is unusable on my HiDPI screen. It shouldn't impact non-HiDPI displays.

See https://doc.qt.io/qt-5/highdpi.html

### DIFF
--- a/muse3/muse/main.cpp
+++ b/muse3/muse/main.cpp
@@ -527,6 +527,7 @@ int main(int argc, char* argv[])
   #endif
 
         // Now create the application, and let Qt remove recognized arguments.
+        QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
         MuseApplication app(argc_copy, argv_copy);
         if(QStyle* def_style = app.style())
         {


terminator356:
Probably the most sweeping comprehensive "one-liner" patch added to MusE so far ;-)
I will be modifying it slightly with an `#if` to check the Qt version first. Stand by...
I don't have such a monitor. Please let me know if I can do anything more.
Thanks so much for researching. I was unaware Qt had this feature, and indeed was
 worrying the last several months about what exactly 'stock' MusE would look like
 (obvious answer: Too small). Worried as well whether our Appearance Settings user-adjustable
 font sizes would be enough to compensate (obvious answer: Meh, it should help, but no).
Perhaps take a snapshot of your MusE in the worst-case highest resolution
 and let me see if anything else needs adjustment.
We usually try to be careful in designing our custom widgets.
From the linked Qt doc:
> In the longer term, the application should be adapted to run unmodified:
>     Always use the qreal versions of the QPainter drawing API.
Well, we have never used the (relatively newer) qreal drawing. We're old, eh !
For myself, I have a fear of trusting floating point coordinates to draw at the
 exact desired location. My experiences weren't good with qreal,
 so I stick with good ol' integers, also our 'View' class has customized
 rounding-off code which didn't mesh well with the qreal floating point.
Using integers and custom rounding we get the exact results we want
 with no misplaced pixels - something that seemed to plague me when I tried qreal.
>     Size windows and dialogs in relation to the screen size.
>     Replace hard-coded sizes in layouts and drawing code by values
>      calculated from font metrics or screen size.
Yes, font metrics and layouts are respected in our custom widgets - we **have** to
 since MusE provides user-adjustable font sizes in the Appearance Settings.
Those settings provide a way for users to custom 'scale' MusE.
However there are a few places where trivial 'margin space' and so on are hard-coded.
I'd love to see what it looks like on HDPI...

Thanks!

PR #635, opened 2019-01-19
Fix build failure

hfiguiere:
I had to do this in order to build because of the build with -Werror.

```
[ 10%] Building C object muse/lv2Support/CMakeFiles/lv2_support.dir/sord-0.14.0/src/sord.o
/home/hub/source/muse/muse3/muse/lv2Support/sord-0.14.0/src/sord.c: In function ‘sord_new_literal_counted’:
/home/hub/source/muse/muse3/muse/lv2Support/sord-0.14.0/src/sord.c:1063:3: error: ‘strncpy’ specified bound 16 equals destination size [-Werror=stringop-truncation]
   strncpy(key.meta.lit.lang, lang, sizeof(key.meta.lit.lang));
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
make[2]: *** [muse/lv2Support/CMakeFiles/lv2_support.dir/build.make:314: muse/lv2Support/CMakeFiles/lv2_support.dir/sord-0.14.0/src/sord.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:2475: muse/lv2Support/CMakeFiles/lv2_support.dir/all] Error 2
make: *** [Makefile:152: all] Error 2
```

### DIFF
--- a/muse3/muse/lv2Support/sord-0.14.0/src/sord.c
+++ b/muse3/muse/lv2Support/sord-0.14.0/src/sord.c
@@ -1060,7 +1060,7 @@ sord_new_literal_counted(SordWorld*     world,
 	key.meta.lit.datatype = sord_node_copy(datatype);
 	memset(key.meta.lit.lang, 0, sizeof(key.meta.lit.lang));
 	if (lang) {
-		strncpy(key.meta.lit.lang, lang, sizeof(key.meta.lit.lang));
+		strncpy(key.meta.lit.lang, lang, sizeof(key.meta.lit.lang) - 1);
 	}
 
 	return sord_insert_node(world, &key, true);


terminator356:
Hi, I too saw this and went deep into the LV2 code to try and solve it
 but I could not really come away with a sense of what the safe correct solution would be.
If I recall I have informed the LV2 author but have not heard back in some time.
I don't think it's been fixed in the latest versions. Still, I need to replace our dated internal version.

I studied adding a solution like this, but I was worried it was not right.
After all, if a solution like this is still not yet upstream we must be careful.
Perhaps best to contact the LV2 author first.
Before I merge this, how sure are you that this is safe for the rest of our internal LV2 version ?
Quite sure? Then I'll do it for now, just to keep us running.
Thanks.

hfiguiere:
The only question is whether they expect more than 15 chars in lang, but then can have more than 16 and in that case the string won't be nul terminated.

PR #665, opened 2019-10-05
Remove all.h that hasn't been used since... so long

hfiguiere:
The rationale for doing it:

it unecessarily complicates the build process, and that somewhat break when building with flatpak and using the standard "cmake-ninja" buildsystem.

From the code history, `all.h` has been empty since before muse3.

Thanks.


terminator356:
In ChangeLog there is this:
> 15.12.2010:
> Major cleanup in the #includes. Emptied all.h and moved its contents to individual files.
> The build is roughly 20% faster now. (Orcan)

So all.h previously contained a list of desired pre-compiled headers.
I have to admit I have never used pch in my own projects and I am
 not sure if, how, and when they are being used in MusE.
So I have always left that stuff alone in MusE.

Also, there are a lot of -fPIC (Position Independent Code) terms.
MusE is about twenty years old now.
As one StackOverflow comment says:

> It used to be the case that when building the shared library, you needed to ensure
>  -fPIC was used for all the object files to be built into the shared library. The rules
>  may have changed because compilers build with PIC code by default, these days.
> So, what was critical 20 years ago, and might have been important 7 years ago,
>  is less important these days, I believe. Addresses outside the o/s kernel are
>  'always' virtual addresses'.

Thank you for doing the heavy lifting here!
Tim.

PR #678, opened 2019-11-09
Color setting for transport handles is ignored in themes

kybos:
Transport handle colour is set to blue in Appearance Settings.
Theme used: Ardour.

Current state:
muse_transport_handle_color_wrong

Fix:
image


terminator356:
Thanks again.

PR #683, opened 2019-11-16
Don't draw invalid piano keys in piano roll

kybos:
When on a small zoom level, the piano used an infinite (tiled) pixmap, so invalid keys below C-2 were visible.

Before:
image

Fix:
image

### DIFF
--- a/muse3/muse/midiedit/piano.cpp
+++ b/muse3/muse/midiedit/piano.cpp
@@ -289,7 +289,15 @@ static const char *mk8_xpm[] = {
     "........................................",
     "........................................",
 };
-      
+
+static const char *mke_xpm[] = {
+    "40 3 1 1",
+    "# c #000000",
+    "########################################",
+    "########################################",
+    "########################################",
+};
+
 /*
       0   1   2  3  4  5  6  7  8  9  10
       c-2 c-1 C0 C1 C2 C3 C4 C5 C6 C7 C8 - G8
@@ -349,6 +357,8 @@ Piano::Piano(QWidget* parent, int ymag, MidiEditor* editor)
       mk6 = new QPixmap(mk6_xpm);
       mk7 = new QPixmap(mk7_xpm);
       mk8 = new QPixmap(mk8_xpm);
+
+      mke = new QPixmap(mke_xpm);
       
       keyDown = -1;
       button = Qt::NoButton;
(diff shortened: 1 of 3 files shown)


terminator356:
Thanks!

terminator356:
OK upon reexamination of photo 2 there is a problem.
I can deduce just from looking at it that the vertical scrollbar is
 **not** setting its range correctly, apparently due to the hidpi.
The vertical scrollbar should **never** allow empty space to show at the bottom.
Because that is both a waste of space, and out of range.
It has never drawn out of range here. I do not have hidpi (yet).
Therefore your fix is a fine safety net, but I believe it should not have been necessary,
 I believe the root problem is the scroll bar range but I can't check it here.
All that stuff is what I just touched when I added persistent editor window settings,
 especially class ScrollScale (our custom vertical scroll and scale control)..
It should be easier to modify now, it wasn't behaving well before.
Watch out if you intend to debug, class ScrollScale may do some re-ranging
 of the scroll part all by itself, upon vertical zoom etc. I think

kybos:
I thought that was so by design.

I assumed that all the zoom levels were fixed, and when the canvas gets higher than the complete piano, then there should be empty space at the bottom if necessary. The scroll bar behaves correctly, it is not displayed when not required, but when I make the windows smaller (so the piano does not fit as a whole), then the scrollbar reappears.

QTractor behaves exactly the same way, by the way. I don't think it's bad, unless it causes issues I am not aware of.

You think the piano should be always scaled to the canvas height as minimum? Would that not cause other issues, like drawing artefacts when resizing the window etc.?

 

 

Gesendet: Montag, 18. November 2019 um 08:59 Uhr
Von: "Tim" <notifications@github.com>
An: muse-sequencer/muse <muse@noreply.github.com>
Cc: kybos <kybos@web.de>, Author <author@noreply.github.com>
Betreff: Re: [muse-sequencer/muse] Don't draw invalid piano keys in piano roll (#683)


OK upon reexamination of photo 2 there is a problem.
I can deduce just from looking at it that the vertical scrollbar is
not setting its range correctly, apparently due to the hidpi.
The vertical scrollbar should never allow empty space to show at the bottom.
Because that is both a waste of space, and out of range.
It has never drawn out of range here. I do not have hidpi (yet).
Therefore your fix is a fine safety net, but I believe it should not have been necessary,
I believe the root problem is the scroll bar range but I can't check it here.
All that stuff is what I just touched when I added persistent editor window settings,
especially class ScrollScale (our custom vertical scroll and scale control)..
It should be easier to modify now, it wasn't behaving well before.
Watch out if you intend to debug, class ScrollScale may do some re-ranging
of the scroll part all by itself, upon vertical zoom etc. I think

terminator356:
Sorry for the noise kybos, I deleted my comment. Please ignore.
Duh. I was way off there. Not thinking.

Issue #638, opened 2019-02-08
Text too small on HiDPI


hfiguiere:
Some of the text is way too small on some UI like "select plugin" dialog.

![image](https://user-images.githubusercontent.com/114441/52459885-af8c9d80-2b35-11e9-902c-f50139649a69.png)

kybos:
I'm just trying to fix that on my local copy of MusE:
![image](https://user-images.githubusercontent.com/10009541/67892034-28180c00-fb54-11e9-8544-9d307965b841.png)
I know of this dialog ("select plugin"), and the Track name change dialog. Are there more of this kind (with too small fonts)?
I will try and learn about that pull request business and then perhaps propose the fixes to the upstream.

kybos:
Fixed in master.

Issue #637, opened 2019-02-08
Cursors are too small on HiDPI


hfiguiere:
Custom cursors are too small on HiDPI. 

See screenshot:

Regular cursor:
![image](https://user-images.githubusercontent.com/114441/52455332-8f50e480-2b1d-11e9-84de-d49be478001c.png)

Custom cursor:
![image](https://user-images.githubusercontent.com/114441/52455337-937d0200-2b1d-11e9-861a-5f259e54bbee.png)

terminator356:
I wonder if the solution is simply to use SVG scalable cursors.
Coincidentally I investigated that during the last big commit. I wondered if it was possible.
Yet I could find nothing about it at the time. But now I found one (!) lonely reference to it
 in all the google results.
So I can try it. If you can wait, I now have access to a 4K monitor and card, just not here.
I might be wrong but the top picture sure does look like an SVG image (smooth).

MusE still uses a lot of bitmaps.
But gradually they are being replaced by SVG versions. (Mixer strips, some others etc.)
For now, I will assume that those particular **bitmap** icons, and the SVG icons, look OK for you.
(It would not surprise me because they're all drawn with QPainter and so on.)

Thanks.

hfiguiere:
Some dialog still use a font that is small. I'll file another issue

PR #684, opened 2019-11-17
Scalable icons/cursors for the edit tools

kybos:
I just wanted to make the tool icons bigger on HiDPI, but it was not as simple as that... So I ended up with all this, as I don't like work left half-done...
I hope my icon creations are acceptable, I strived to make them look OK on all three default themes.


kybos:
It also solves this issue: #637 
You can also close this one: #638 (resolved by an earlier PR)

terminator356:
Oh boy! Let's see what they look like... Thanks!

terminator356:
Hi, I noticed a few things.
`getCursorFromIcon(const QIcon* icon, const QString& iconname) {`
`const qreal dpr = qApp->devicePixelRatio();`
`int hotX, hotY;`
`if (iconname == "pencilIconSVG"`
Can we get away with just comparing the pointers to the icon variables instead of strings?
After all, the strings are exactly the same as the SVG icon variable names.
ie. eliminate iconname.

Also, is it the case where you are expecting devicePixelRatio() to change
 dynamically at run time? If not, I'd suggest creating the cursors once at init,
 just like editpasteSCursor and editpasteCloneSCursor, which are the two cursors
 found in icons.cpp. I worry that dynamically creating the cursors each time
 the cursor needs to change might be a bit heavy? I'm thinking that when we use
 the already-created Qt built-in cursors or our own ones, the switch is faster?

Last, you need a 'closed hand' to **compliment** your 'open hand',
 for (at least) this section in Canvas::setCursor():
`        case DRAG_PAN:`
`        if(MusEGlobal::config.borderlessMouse)`
`            showCursor(false); // CAUTION`
`        else`
`            QWidget::setCursor(QCursor(Qt::ClosedHandCursor));`
`        break;`

Thanks.

kybos:
1. Yes, you are absolutely right with just comparing icon pointers. I forgot that I can rely on the icons having been initialized when the function is first called.

 

2. I'm afraid devicePixelRatio() can change dynamically indeed. Imagine you have two monitors, one low-res, one HiDPI and you move the application between them.

But that is a special use case and most applications struggle with it badly, even commercial ones. For the sake of efficiency and consistency, I will do as you suggest.

 

3. Closed hand cursor: I will look into it. I would have preferred to use Qt's own hand-cursors, but for some reason they do not scale on HiDPI. I suspect that they have no pendants in the icon theme on Linux so Qt' fallback cursors are used instead, which apparently do not scale correctly.

 

I will supply a PR when ready.

 

 

Gesendet: Montag, 18. November 2019 um 08:08 Uhr
Von: "Tim" <notifications@github.com>
An: muse-sequencer/muse <muse@noreply.github.com>
Cc: kybos <kybos@web.de>, Author <author@noreply.github.com>
Betreff: Re: [muse-sequencer/muse] Scalable icons/cursors for the edit tools (#684)


Hi, I noticed a few things.
getCursorFromIcon(const QIcon* icon, const QString& iconname) {
const qreal dpr = qApp->devicePixelRatio();
int hotX, hotY;
if (iconname == "pencilIconSVG"
Can we get away with just comparing the pointers to the icon variables instead of strings?
After all, the strings are exactly the same as the SVG icon variable names.
ie. eliminate iconname.

Also, is it the case where you are expecting devicePixelRatio() to change
dynamically at run time? If not, I'd suggest creating the cursors once at init,
just like editpasteSCursor and editpasteCloneSCursor, which are the two cursors
found in icons.cpp. I worry that dynamically creating the cursors each time
the cursor needs to change might be a bit heavy? I'm thinking that when we use
the already-created Qt built-in cursors or our own ones, the switch is faster?

Last, you need a 'closed hand' to compliment your 'open hand',
for (at least) this section in Canvas::setCursor():
case DRAG_PAN:
if(MusEGlobal::config.borderlessMouse)
showCursor(false); // CAUTION
else
QWidget::setCursor(QCursor(Qt::ClosedHandCursor));
break;

Thanks.

kybos:
Re: Deleted pixmaps:
I may be a cleanup fanatic, but I frankly don't see a point in crowding current code line with obsolete objects. We are on git after all, nothing is ever really deleted, you can always find your objects if you want them.
But you are the boss here, I won't delete anything, I promise ;-).

terminator356:
Oh, sorry I meant the actual xpm files, not the code.
The code can be completely removed of them if you like. Absolutely.
Just the actual xpm files I'd like to keep around if possible.
They're small and as I say, sometimes I use them for inspiration when creating new SVGs.
Perhaps we could do a round up of unused ones and dump them in a tar file or something.

For number 2: OK I see. Go ahead and leave that in then, that's fine if you like.
If it makes things better, no problem ;-)

terminator356:
>I'm not sure that now is a good time to be debating whether to
clean house of them all...

Duh. Like right now when someone is actually making SVGs...

>Perhaps we could do a round up of unused ones and dump them in a tar file or something.

Duh. Like old git objects...
Honestly my only though there was "what if github goes away?" or something like that.
Just didn't want to lose them.

You'll have to excuse my 3 AM posts sometimes...
Please, carry on. Thanks.

kybos:
When github goes away, that would be kind of end of the world, I think ;-). But still: We all have 1:1 local copies of the whole repo, so even then nothing would get lost in fact. 
You can for example checkout to a previous release tag to retrieve the old files, or simply switch to it on github web.
I am not trying to convince you, just giving my personal opinion. I just prefer to remove everything that is not used from the current code, that keeps the setup clean and less confusing for new users ;-).

terminator356:
No prob. You'll see in a few places large commented sections.
Or sections that are ifdef 0 'd out.
Sometimes they are (were) coders' experiments that were alternate or 'what if' scenarios.
For my own such sections, sometimes I'm reluctant to remove them until certain.
Later as I pass them again I try to clean up.
Especially these days I mark my experiments with "REMOVE Tim..."
 and try to make sure I clean up when the experiment is confirmed or not.

One such ugly large commented area is in the wave track module.
They were early attempts to support live re-sampling and time-stretching.
I have a **very** highly advanced branch that includes all of that !
I'm trying very hard to catch up to it so I can merge it into master.
It keeps slipping away, so I want to merge what I've got as long as it
 doesn't affect anything when it's disabled.
I'm tired of having to always merge master into it, with all the conflicts.
Just get the darn thing into master and work on it there!

terminator356:
Around the same time Florian made a fork with similar time-stretching features
 except he added beat detection and matching, something I'm still in planning stages with.
We talked about it and agreed that knowing the tempo map of an imported sample is crucial.
It has to do with how it fits into the song and having a tempo map to refer to against our own,
 to manipulate the sample when stretching.

kybos:
Yes, I understand leaving comments of this kind in code for some time, that's OK of course.
But it would be nice to clean up the folder with the xpm files (and icon.cpp/h if there are unused objects, too), to have a clear picture of how many are actively in use and how many still remain to be converted to SVG, for example. Their current number is quite intimidating ;-).

PR #687, opened 2019-11-17
Restrict note tooltip to editing tools

kybos:
Improvement for #673.
Don't show the tooltips when non-editing tools like zoom, pan are active.


terminator356:
Et merci!

Issue #686, opened 2019-11-17
Master track graphical editor issues


kybos:
I discovered several issues with this editor while working at my last PR:

1. Hard-coded background colour, not theme-aware.
2. Tempo and signature stripes do not show anything when not on default theme (see screenshot).
3. Draw tool: Drawn line is not visible at all (at least not on HiDPI screen).
4. Draw tool does not work backwards (when line is drawn from right to left).
5. The grid lines: The beat lines seem wider and thus more accentuated then the bar lines, which is a little confusing.

Probably not so critical: 
6. No right-click menu, no working keyboard shortcuts for scrolling/zooming etc.

![image](https://user-images.githubusercontent.com/10009541/69007717-9b39c480-0941-11ea-85d6-a80d54d22768.png)

kybos:
By the way: If nobody is working on that, perhaps I could try to help out (at least with the UI issues).

spamatica:
There are many things to work on in MusE ;) I think I can safely say no one
is working on the UI issues at the moment.

If you feel inspired you should definitely go for it :)


Den sön 17 nov. 2019 kl 14:48 skrev kybos <notifications@github.com>:

kybos:
OK, I think I fixed 1-5.
PR will follow.

![image](https://user-images.githubusercontent.com/10009541/69010649-fc719000-0961-11ea-8a4d-7c76b506ca7b.png)

terminator356:
I'm to blame for virtually all coding done here in the last several years ;-)
This editor is neglected and lags more than the others, but I made many important drawing
 fixes a few years back. (Tip: Use 'cosmetic' pens when drawing single-pixel things
 otherwise lines will disappear.)

Hm yeah, backward line drawing messed up. Guess I didn't fix that.
Line barely visible light grey colour here. How'd that happen.

Because of the similarities with the midi controller graph editor, which I really souped up recently,
 I was gonna sort of, kind of, try to modernize it with those same copy/cut/paste/drag features,
 but didn't get around to it. Alas, it's lower priority than say, engines and plugins! So many things...

terminator356:
Use your zen-like powers to concentrate.
Tempos and time signatures are stored in mind-warping way ;-)
There's always at least one tempo and time sig and they're stored at tick number gazillion,
 after all the other entries, I think. It's kinda weird but necessary, esp with the drawing.

kybos:
> I'm to blame for virtually all coding done here in the last several years ;-)
[...]

PR #688, opened 2019-11-17
Master track graph editor ui issues

kybos:
See  #686


kybos:
My mistake, to be redone...

PR #689, opened 2019-11-17
Fix master track UI issues

kybos:
See #686 for details.

Please check especially the change in view.cpp.
If I understand the code correctly, there were 2 parallel lines drawn for each grid beat line. I don't know why, it certainly does not look good on my screen, but I guess there was a reason for that?
I just don't want to sabotage Tim's work ;-) (it was him according to git blame).

### DIFF
--- a/muse3/muse/components/sigscale.cpp
+++ b/muse3/muse/components/sigscale.cpp
@@ -141,6 +141,7 @@ void SigScale::pdraw(QPainter& p, const QRect& r, const QRegion&)
       if (x < 0)
             x = 0;
       p.setFont(MusEGlobal::config.fonts[3]);
+      p.setPen(Qt::black);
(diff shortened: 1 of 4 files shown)


terminator356:
Actually those double line draws are Robert's commits according to my KDevelop annotations.
I think I remember seeing the commits but was puzzled what the description meant.
Robert anything to add? Did it help visibility there?

terminator356:
Ha! Yep, cosmetic pens. Upon looking at the pull request I'm fairly certain
 that would have been Robert's problem. No cosmetic pen. Geez, how'd I miss that.
Hang in there... Must get merge order correct... Thanks.

spamatica:
Ah, oh. I didn't release it was the gridlines you were talking about. 
Yeah, the double lines were a test to get the beats more visible. Can't remember now, thought I scrapped it but maybe not. Anyways, if it only causes problems it shouldn't be there.

PR #696, opened 2019-11-21
Thicker grid bar lines/fix grid update in ctrl

kybos:
I was trying to solve the problem of the 3 kinds of (vertical) grid lines not being clearly distinguishable (bar, beat and sub-beat, relevant for pianoroll and drum edit). So I made the bar line a little thicker, and the beat line black in my settings, now it seems much clearer to me (there is no customization option for the sub-beat lines, unfortunately).
Please check if it looks OK for you. I am not sure about a low-resolution screen, as  I don't have any.

image

I noticed another issue: When the raster was changed (from the snap dropdown on the toolbar), then only the pianoroll/drum canvas was updated, but not the control area(s) below it. This is also fixed here.

### DIFF
--- a/muse3/muse/components/view.cpp
+++ b/muse3/muse/components/view.cpp
@@ -951,6 +951,7 @@ void View::drawTickRaster(
 
                     //>pen.setColor(bar_color);
                   }
+                  pen.setWidth(2);
                   p.setPen(pen);
                   
                   if(scale_info._drawBar)
@@ -970,6 +971,8 @@ void View::drawTickRaster(
                       p.drawLine(mx_sm, my, mx_sm, mbottom);
                     }
                   }
+                  pen.setWidth(1); // reset
+                  p.setPen(pen);
                   
                   if(drawText)
                   {
--- a/muse3/muse/ctrl/ctrledit.h
+++ b/muse3/muse/ctrl/ctrledit.h
@@ -83,6 +83,7 @@ class CtrlEdit : public QWidget {
       // Special: We 'abuse' a controller event's length, normally 0, to indicate visual item length.
       void tagItems(MusECore::TagEventList* tag_list, const MusECore::EventTagOptionsStruct& options) const
       { if(canvas) canvas->tagItems(tag_list, options); }
+      void redrawCanvas() {canvas->redraw();}
       };
 
       
--- a/muse3/muse/midiedit/drumedit.cpp
+++ b/muse3/muse/midiedit/drumedit.cpp
@@ -1071,6 +1071,8 @@ void DrumEdit::setRaster(int val)
       _rasterInit = val;
       MidiEditor::setRaster(val);
       canvas->redrawGrid();
+      for (auto it : ctrlEditList)
+          it->redrawCanvas();
       focusCanvas();     // give back focus after kb input
       }
 
--- a/muse3/muse/midiedit/pianoroll.cpp
+++ b/muse3/muse/midiedit/pianoroll.cpp
@@ -1306,6 +1306,8 @@ void PianoRoll::setRaster(int val)
       _rasterInit = val;
       MidiEditor::setRaster(val);
       canvas->redrawGrid();
+      for (auto it : ctrlEditList)
+          it->redrawCanvas();
       focusCanvas();     // give back focus after kb input
       }
 


terminator356:
Thanks. I'll check that line thickness...

spamatica:
Nice, will check how it looks here. 
Tim, it's already merged but for the record it's possible to checkout a pull request into a separate branch to test before merging.

git fetch origin pull/<id of the pull request>/head:<the new branch> 
git checkout <the new branch>
build and test

terminator356:
Oh thanks! I was wondering if that was possible.

Kybos, I know this sounds like an insanely minute detail, but 2 pixels wide is
 not really a good idea. What is the nature of the problem you're observing?
Is it a case where the bar lines are indistinguishable from beat lines?
It would be better to try and solve that with say, colouring.
There may be competition among three lines for the space but ultimately
 the bar line must always win and always be drawn on top of any other lines that are
 trying to be drawn there. I know there may be some problems with who wins there,
 I mentioned it was not an exact science and I recall having trouble in some areas.

The problem with 2 pixels is that one cannot tell where centre should be.
If it becomes absolutely necessary to draw a line thicker than 1 pixel,
 one must jump the next odd value which is 3. At that point the line becomes pretty thick.
The only time I reluctantly did that is with the 'velocity spikes' on the velocity graphs.
2 pixels was impossible to work with, so I had to make it 3, centered in the middle.

Another exception is the edges of parts. That's another story, a careful attempt
 to make the butts of parts mesh well, especially since one or the other can be
 a clone part's dashed lines meshing with solid lines.

Thanks.

spamatica:
eh. my git command examples were completely borked. seems the comment field strips away what looks like tags as I used that format
It should have said something like:
git fetch origin pull/ID/head:BRANCHNAME
where ID is the pull request id and BRANCHNAME is the new branch name.
and
git checkout BRANCHNAME
to switch to the branch
More info here:
https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally

terminator356:
Ha! I thought that looked suspicious...

spamatica:
> The problem with 2 pixels is that one cannot tell where centre should be.

I can understand this from a technical point of view. There will be an offset so it's not 'right', but do we really have somewhere in the application where it matters? If you adjust stuff without snap maybe? (But then you are on thin ice anyway...)
This was the effect I was after with my double-line fix from before, though mine didn't work that well.

terminator356:
Yes, here, because the line needs to be centered on the exact (abstract) time line.
That's why I drew the velocity spikes 3 pixels so that they are centered on the graph line.

kybos:
Perhaps I am on the wrong track here, but I think it's just the opposite - you cannot draw a 1-pixel line exactly on the coordinate point, in fact it's drawn to the right of this (abstract) point (in Qt at least). A line with even pixel number can be drawn symmetrically to both sides of the point, so it's visually more correct.
See here:
https://doc.qt.io/qt-5/coordsys.html

terminator356:
See how 2 pixels grid doesn't quite jive with the small black mouse line which
 is the real representation of the time line:
vertical_lines_1

terminator356:
@kybos That's something different. Don't worry we deal with all of that Qt offset stuff.
One-pixel grid lines are in fact representing our time lines as exactly as possible,
 they are 'bang-on'.

terminator356:
@kybos Oh I see, do you mean use anti-aliasing to draw 2 pixels to
 look like a centered line? Isn't that still using 3 or 4 pixels though?
If so, a price to pay with anti-aliasing is fuzziness of the lines.

kybos:
No, no anti-aliasing, that would make no sense at all for vertical lines (and I hate it, it makes my eyes sore, also the reason why I got HiDPI screens as early as sensibly possible).
No problem, if it collides with the existing system, I will revert the change. On my monitor it looks perfectly aligned, but I know that normal HD monitors are different, that's why I asked you to check in the first place.

terminator356:
Different how? They can draw a 2 pixel line centered?
I really must get my hands on one...

terminator356:
BTW If you're wondering why we didn't just switch over to Qt's floating point
 coordinate system when it became available in Qt4, I had some trouble with it.
Floating point caused grief with accuracy. You'll see a custom integer system in there
 with some hopefully helpful comments.
It's the same reason I switched all our critical timing variables from floating point 
 over to large integer math. It really helped, timing was not accurate before.

kybos:
Don't know, but for me it look OK:
image

terminator356:
No no, see my picture, the small black line in the timescale **above** the numbers.
It follows your mouse around. Hard to notice sometimes.

kybos:
Yes, I know, it's there on my picture, light grey, above the 17... It's the Ardour theme ;-)

terminator356:
Oh wait, it's there. I see it. (Ardour theme...)

terminator356:
In my pic the difference is more  pronounced black lines one white bkg.
I guess not being hidpi doesn't help either.

spamatica:
After some thought I have realized that Tim is right. The lines should not be two pixels wide. Some more thought and I realized that the probable reason it was hard to get good visibility of the different lines was that some colors had hard-coded adjustment, e.g. color.lighter(200);
I made a new pull request https://github.com/muse-sequencer/muse/pull/699
 with some rework to get more colors editable in the appearance dialog.
Do test and see if it looks okay on your end, seems to work very well here. (unless I mixed up any of the colors, had to try too many times before I got it to where I wanted)

PR #698, opened 2019-11-21
Revert thick bar lines

kybos:
Refer to #696

### DIFF
--- a/muse3/muse/components/view.cpp
+++ b/muse3/muse/components/view.cpp
@@ -951,7 +951,6 @@ void View::drawTickRaster(
 
                     //>pen.setColor(bar_color);
                   }
-                  pen.setWidth(2);
                   p.setPen(pen);
                   
                   if(scale_info._drawBar)
@@ -971,8 +970,6 @@ void View::drawTickRaster(
                       p.drawLine(mx_sm, my, mx_sm, mbottom);
                     }
                   }
-                  pen.setWidth(1); // reset
-                  p.setPen(pen);
                   
                   if(drawText)
                   {


terminator356:
Thanks. It's all good.

kybos:
Not quite good ;-), I still think there is a usability issue with the grid, especially on the darker themes. I find it difficult to make the three types of line clearly discernible, at least when using only shades of grey. Trying to find a way without making the lines thicker...

One problem is the colour of the sub-beat lines being hard-coded (Qt::darkGray), that restricts the customizing possibilities considerably.

So here is another suggestion:
1. Make the colour of the sub-beat lines customizable (in Appearance settings), like the bar and beat line.
2. Optional: Make the sub-beat lines dotted (looks quite cool on my monitor ;-), and makes the lines automatically appear lighter in colour).

What do you think?

terminator356:
1. No problem, the more customizable the better.
Feel free to add whatever is missing, + any other lines. Not quite the most trivial thing though:
The colours are an array. Add to it. Add to the conf write/read. Add to the colour editor.

2. Ha! Dotted lines. In my time-stretching branch I had to resort to using them.
Several line types and colours already exist - markers, cursors, grids etc -
 so I needed two new line types for movable 'stretch markers'. I had to use two different
 coloured dotted lines. Was never quite pleased with it.
Whether it'd work here in the grid, I'm not sure. I prefer a solid, yet unobtrusive grid,
 but I have seen grids with dotted lines (LibreOffice etc.). I... could... be convinced...

spamatica:
Hehe, didn't read the messages here... seems you came to the same
conclusion Tim. Here's hoping there isn't too much duplicate effort ;) See
the message I wrote in the other ticket and the related pull request.


Den fre 22 nov. 2019 kl 17:04 skrev Tim <notifications@github.com>:

PR #699, opened 2019-11-22
added more editable colors used on canvases

spamatica:
bild

Took another stab at fixing visibility of lines on canvases. Basically added more editable colors and removed more of the hard-coded colors, as I now think that was the main reason it was hard to get a good mix.

### DIFF
--- a/muse3/ChangeLog
+++ b/muse3/ChangeLog
@@ -1,3 +1,6 @@
+22.11.2019
+    - New solution to making bar lines more visible. Basically removed more hardcoded colors
+      and made them configurable in the appearance dialog (rj)
 21.11.2019
     - Add command line option -C "Do not re-create plugin cache." (Tim)
       Avoids (looong) repeated re-creations in some cases. Helps devs avoid after rebuilds. 
--- a/muse3/muse/appearance.cpp
+++ b/muse3/muse/appearance.cpp
@@ -186,6 +186,9 @@ Appearance::Appearance(Arranger* a, QWidget* parent)
              new IdListViewItem(0x600 + i, id, MusEGlobal::config.partColorNames[i]);
            
            new IdListViewItem(0x41c, aid, "part canvas background");
+           new IdListViewItem(0x42c, aid, "part canvas raster coarse");
+           new IdListViewItem(0x42d, aid, "part canvas raster fine");
+
            new IdListViewItem(0x41f, aid, "Ruler background");
            new IdListViewItem(0x420, aid, "Ruler text");
            new IdListViewItem(0x424, aid, "Ruler current marker space");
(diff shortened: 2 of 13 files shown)


spamatica:
Just noticed the velocity editor now has Cyan colored lines. This is due to me changing the default color (to catch this kind of thing, uninitialized colors)

kybos:
Oh, looks like we were working on the same thing...
I guess you can ignore my current PR then...

terminator356:
Couple of notes:
1: If you're wondering about the blue coloured master track graph,
 I changed that in my other branch to the same colour as the controller graphs (default orange-ish).
I could cherry-pick that to get it in sooner.

2: The way we set colours with our style sheets is brutal and destructive.
We have an extra accompanying file containing colours which **overwrite** 
 the user's configuration colours, and is stored that way when MusE closes.
This may have been added before Qt allowed access to object 'properties' from stylesheets:
> From 4.3 and above, any designable Q_PROPERTY can be set using the
 qproperty-<property name> syntax.
For example,
  MyLabel { qproperty-pixmap: url(pixmap.png); }
  MyGroupBox { qproperty-titleColor: rgb(100, 200, 100); }
  QPushButton { qproperty-iconSize: 20px 20px; }
If the property references an enum declared with Q_ENUMS, you should
 reference its constants by name, i.e., not their numeric value.

I propose adding setter and getter functions together with properties declarations,
 so that the colours can be **directly accessed from the stylesheet**.
Which class to put the properties, not sure, likely in class View.
However, the getters and setters must **not** directly access and destroy the user's
 configured colours. We'll have to keep a separate '**working copy**' of the colours.
The working copy is loaded with the user's configuration colours at startup,
 and any further stylesheet loaded will overwrite **these** working colours.

Thus all places which use the user's configured colours will have to use
 the 'working copy' instead.
But it's a much cleaner solution than having a separate accompanying 'colour file'.

If we're lucky, we may not need separate getters/setters for each colour.
We may be able to use a single getter/setter with an index parameter.
At first this sounds not good, how does the author know which index
 corresponds to which colour?
But this statement is very encouraging, it means we could use enum'd colour names
 instead of an index number:
> If the property references an enum declared with Q_ENUMS, you should
 reference its constants by **name**, i.e., not their numeric value.

spamatica:
1. Got it!
2. I sorta see where you are going, this seems a little bigger than I had in mind for the moment though.. or, I don't know how this works, maybe it's not that big.

kybos:
Just tested your changes.
That solves my problem with the grid lines perfectly, thanks.

terminator356:
Hi Robert. I get this when snap is on 8 or 16.
You've got two uninitialized 'reds' in there feeding drawTickRaster(),
 designed as red to shout out that they are uninitialized.
Was there meant to be two additional customizable colours added, to feed in there? Thanks.
red lines 1

spamatica:
What do you know, I missed more of them. Yeah, I added a few more settable
colors and apparently they should be used here too. I'll fix it tomorrow.

/Robert

Den lör 23 nov. 2019 kl 18:54 skrev Tim <notifications@github.com>:

PR #700, opened 2019-11-22
New setting for subbeat grid lines

kybos:
Refer to #698
No need for dotted lines now. From the normal working distance they look as light solid lines anyway...


spamatica:
lol, sorry. 
I'll check later how much duplicate effort I caused...

kybos:
No problem, your PR is much more comprehensible, I was only after the sub-beat lines...

Issue #669, opened 2019-10-31
Advanced routing dialog with Ardour theme (unreadable fonts)


kybos:
The dialog looks quite bad with the Ardour theme:

![image](https://user-images.githubusercontent.com/10009541/67971363-f0b96600-fc0c-11e9-92a6-d8097aa18409.png)

It can be made a lot better adding this to Ardour.qss:

 QTreeView {
    alternate-background-color: rgb(127,127,137);
}

Then it looks like this:

![image](https://user-images.githubusercontent.com/10009541/67971750-a1276a00-fc0d-11e9-980f-8c98567f0146.png)

But I am not able to change the section titles colour (like "Jack:", "Midi Ports:", "Tracks:" etc.) - these texts are still unreadable. Can you help?

Also, could the dialog be made a little larger as a default (or the size kept between sessions)?

Thanks,
Kybos

kybos:
Addition: This should be added to Ardour.qss:

QTreeView, QListView {
    alternate-background-color: rgb(127,127,137);
}

Because also the lists are affected by this issue.

terminator356:
Thanks. Some of our drawing in MusE uses custom painting with
 colours not easily found in a palette or in standard stylesheet syntax.
I will try to expose a new property for that particular colour.
Also must ensure this particular TreeView control has an object name.
That way we can use the 'object property' feature of stylesheets (since Qt 4.3).
There will certainly be other MusE widgets that will need to be done as well.

Please hang in there, quite busy at the moment...

terminator356:
Heya! Got some help for you.
Pardon me if I reopen this just for wee while so we can remember it.

Advanced router stylesheet support: New property 'RouteTreeWidget::categoryColor'
You can now set that 'category' item colour.

**Did you know...**
You can access any specific MusE sub classes of QWidget.
All of our custom widgets are housed in a C++ namespace "MusEGui".
Therefore to access any of our widgets, you must use "MusEGui--<classname>"

Set this new property like this:

`MusEGui--RouteTreeWidget {`
`      qproperty-categoryColor: blue;`
`      background-color: green;`
`      color: cyan;`
`      alternate-background-color: orange;`
`     ## etc etc.
`}`

In addition the two specific instances of that class use now have IDs:
"Router_input_tree" and "Router_output_tree".
You're _supposed_ to be able to access them like this, but I couldn't get it to work:
`MusEGui--RouteTreeWidget#Router_input_tree {`

Cheers!

[...]

PR #702, opened 2019-11-24
Fix theming of toolbutton with submenu

kybos:
Before:
image

Fix:
image

Also added fixes for "category colour", see #669.

### DIFF
--- a/muse3/share/themes/Ardour.qss
+++ b/muse3/share/themes/Ardour.qss
@@ -52,6 +52,14 @@ QToolBar QToolButton:checked {
 
 }
 
+QToolBar QToolButton[popupMode="1"] {
+      padding-right: 12px;
+}
+
 QTreeView, QListView {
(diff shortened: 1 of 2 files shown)


terminator356:
Just to warn in case of head-scratching:
Somewhere in the code I believe there's a line where we force most of the
 toolbar icons to be a certain size. The line may or may not be active these days.
It was done because otherwise the icons and toolbars ended up being too big
 and we were trying to shrink them to be a reasonable size.
I think it may have been because some icons want to be their own natural size which was too big.

kybos:
Yes, I was just going to file an issue about that.
There is a default size of 18 defined, I think in gui.h, which is considered in most places, but not everywhere.
In the non-native plugin UIs the toolbar icons are much too big:
image.

terminator356:
Ah yes. gui.h I completely forgot that file even existed.
Such a lonely define in there all by itself...

PR #705, opened 2019-11-25
Consistent metronome icon (fixes #704)

### LINKED ISSUE #704: Metronome icon inconsistent
I was thinking about replacing the pixelated metronome icon on the main toolbar (and also the other one in the Settings menu) by an SVG one, but at some point I noticed that there already exists an SVG version of it, but it is only used in the transport window. Is that intended? And is it really necessary to have a different metronome icon for the settings menu? It looks more like an inconsistency from user perspective, I think.
I would suggest to replace all these old pixmap icons by the "metronome on" SVG icon that is used in the transport.

kybos:
Not sure what the result of our last discussion actually was:
Can I remove the xpm files for the obsolete metronome icons from master?


terminator356:
Yes ;-)

PR #710, opened 2019-11-30
Native plugin UI - HiDPI scaling (fixes #697)

### LINKED ISSUE #697: Soft synths and HiDPI
Most of the native plugin GUIs open a window that is much too big, normally twice the height/width of the real GUI. I suppose this is caused by Qt scaling the complete UI for HiDPI, so it expects that the content of all windows will be scaled as well. This is even the case with Helm, despite its scaling capabilities.
With Calf plugins the problem is a little different, they open a huge window over the whole screen, which they do fill out, but the content is scaled in a weird way. It takes a couple of seconds before the window comes up.
I assume this is a difficult topic, Ardour and QTractor show similar problems (and more crashes for that matter), only Carla gets it right in most cases.

image

image

kybos:
New global setting created for reverting the HiDPI scaling of incompatible plugins.
Can be overridden per soft synth track in the quirks.

### DIFF
--- a/muse3/ChangeLog
+++ b/muse3/ChangeLog
@@ -1,3 +1,9 @@
+30.11.2019
+	- Fix #697: On HiDPI displays, plugin native UI windows are automatically
+	  scaled by Qt, but most of the native plugin UIs are not scalable 
+	  or apply their own scaling. There is now a global setting for reverting 
+	  the Qt window scaling. It can be overridden for each soft synth track
+	  by a local setting in the Quirks toolbar. (kybos)
 28.11.2019
     - Fixed the DidYouKnow menu alternative so it shows the dialog even if the
       checkbox for not displaying it is checked. (rj)
--- a/muse3/muse/components/genset.cpp
+++ b/muse3/muse/components/genset.cpp
@@ -223,6 +223,7 @@ void GlobalSettingsConfig::updateSettings()
       denormalCheckBox->setChecked(MusEGlobal::config.useDenormalBias);
       outputLimiterCheckBox->setChecked(MusEGlobal::config.useOutputLimiter);
       vstInPlaceCheckBox->setChecked(MusEGlobal::config.vstInPlace);
+      revertPluginNativeGUIScalingCheckBox->setChecked(MusEGlobal::config.noPluginScaling);
 
       deviceAudioBackendComboBox->setCurrentIndex(MusEGlobal::config.deviceAudioBackend);
 
@@ -448,6 +449,7 @@ void GlobalSettingsConfig::apply()
       MusEGlobal::config.smartFocus = smartFocusCheckBox->isChecked();
       MusEGlobal::config.borderlessMouse = borderlessMouseCheckBox->isChecked();
       MusEGlobal::config.velocityPerNote = velocityPerNoteCheckBox->isChecked();
+      MusEGlobal::config.noPluginScaling = revertPluginNativeGUIScalingCheckBox->isChecked();
 
       MusEGlobal::config.addHiddenTracks = addHiddenCheckBox->isChecked();
       MusEGlobal::config.unhideTracks = unhideTracksCheckBox->isChecked();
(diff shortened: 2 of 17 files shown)


terminator356:
And another badge of honour, this time to you for navigating these waters.
I know it is tough. Sheer determination wins. Rest well !

terminator356:
Got this: I'll see if I can fix it 'cause I have to merge my current work soon:

/home/me/Source/muse_20190919_master/muse3/muse/plugin.cpp: In member function ‘void MusEGui::PluginGui::fixNativeUIScalingTBClicked()’:
/home/me/Source/muse_20190919_master/muse3/muse/plugin.cpp:4376:11: error: operation on ‘state’ may be undefined [-Werror=sequence-point]
 4376 |     state = (state == 2) ? 0 : ++state;
      |     ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~

terminator356:
I fixed it by putting state + 1 instead of ++state. I will commit later but feel free
 to go ahead and commit a fix if desired ;-)

About class IconButton: Say, that's very clever use of tr-state!
Perhaps a burden with a two-state button?
I say that because ironically at this very moment I must contemplate making them
 drop-down capable (like the panic button), because I need to add such a
 multi-state button to the transport where space is limited.
But it is not based on abstract button just widget, so I must figure out something
 to allow them to be added to a button group like the panic button.
If I inherit abstract button, I end up with some redundant unwanted inherited things...

terminator356:
Actually, button schmutton. I didn't want an actual button for space reasons,
 that's why I made them like that. Also the 'arrow' would awkwardly take up space.
Your idea however, seems better for a limited numbers of states, say three, four at most?
With a little mods, we can do that to IconButton... maybe not so easy since
 state can't be set with a simple boolean anymore, but hey, why not.

terminator356:
D'oh. Sorry, I'm confused (1:30PM too early for me), the quirk buttons are actual buttons.
I was referring to my buttons on the transport and elsewhere.

kybos:
> Got this: I'll see if I can fix it 'cause I have to merge my current work soon:
> 
> /home/me/Source/muse_20190919_master/muse3/muse/plugin.cpp: In member function ‘void MusEGui::PluginGui::fixNativeUIScalingTBClicked()’:
> /home/me/Source/muse_20190919_master/muse3/muse/plugin.cpp:4376:11: error: operation on ‘state’ may be undefined [-Werror=sequence-point]
> 4376 | state = (state == 2) ? 0 : ++state;
> | ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~

I did not get this error and was able to build, and the tri-state button worked, too...
I guess that's because of the build configuration, I use the "Release with debug info" one.
I know I should use DEBUG because of the stricter checks, and I also tried to, but it always aborts because of other errors (-Wdeprecated-declarations).

kybos:
OK, I added -Wno-deprecated-declarations to my DEBUG configuration, now I can reproduce the error. My "state"-construct was somewhat unclean indeed (even if it worked), your correction is OK.

Still, should not the DEBUG configuration compile with the default settings?

spamatica:
Den sön 1 dec. 2019 kl 12:14 skrev kybos <notifications@github.com>:

> OK, I added -Wno-deprecated-declarations to my DEBUG configuration, now I
> can reproduce the error. My "state"-construct was somewhat unclean indeed
> (even if it worked), your correction is OK.
>
> Still, should not the DEBUG configuration compile with the default
> settings?
>
I believe it was Andrew who added the extra -W flags. At least in theory,
though it is a hassle, I find it a good idea to be extra strict on debug
builds (in the sense that debug is mostly built by developers) to be forced
to fix these "potential" errors, hopefully before they become a problem for
others.

kybos:
Yes, I absolutely agree with that, that was exactly my point: I would like to use the debug configuration, but it does not compile with the current code because of -Wdeprecated-declarations errors.

terminator356:
I rarely build as release.
Therefore in debug mode I catch every warning being treated as an error.
I'm fine with that, I'm used to it by now.
But, yes, it did cause big problems with things out of our control. Header files with warnings etc.
So much so that I had to make a couple of our modules separate libraries just so that
 they could be built without the -w flag.The LV2 module and the LV2 Gtk module
 are examples where I had to do that because of header warnings..
I tried very hard, but could not find a way to disable the -w for specific headers or modules.

PR #711, opened 2019-12-02
New svg icons, UI text capitalization fixes, cleanup

kybos:
1. More SVG replacements for pixelated icons (mostly similar to the original ones).
2. Fixed many UI texts inconsistencies:
 - Approx. 80% of the menu texts used title-style capitalization (like "Open in New Window"), which is good and standard, so I corrected the rest.
 - Other UI texts (except Titles/Menus) are supposed to use sentence-style capitalization (like "Open in new window"). In MusE they were quite mixed, I tried to fix what I could find. Many texts were even completely in lowercase, which, from what I understand, is quite against UX rules.
 - It's a convention to end menu entries that trigger another window/popup with three dots (...) to make this clear to the user (like Save As...). Easy for the developer and very useful to the user ;-). I tried to adjust the relevant entries.
 - Top menu entries with more than one word can be very confusing and hard to distinguish from the rest (depending on desktop environment/theme/display manager). I suggest to contract "Window Config" menu to "WinConfig" (if you mind or have any better idea, I will of course revert this change).
3. Further cleanup of unused icons.


terminator356:
[ Scratches head... ] Hm, this is a puzzling one for me.
I was always taught that you don't capitalize secondary words unless they are proper names.
Thus, our arranger has a proper name "the Arranger" so a (bogus) menu title like
 "Open in Arranger" makes sense. But we have no component with a proper name
 "New Window" so " Open in New Window" falsely implies that there is a component
 somewhere with that proper name "New Window".

However, upon examination of other apps, I don't think I ever noticed before,
 but capitalization seems to be the convention.
Blimey. I suppose the reason could be that
 **the text of the action, not of what it points to, is a proper name, in this case
 the text is the proper name of the action itself**.

Merging. Thanks.

terminator356:
We are going to have to put a call out to translators.
Some of the translations are gone now.

terminator356:
I've never worked with the translations before.
I am wondering why they would be lost.
I am hoping that they are still there and maybe we just regenerate them.
Hopefully they point to the actual objects not the texts themselves? Or no? (Please be yes!)
Florian used to take care of that. I think he added a special cmake option
 for some sort of regeneration that maybe we're supposed to use?
I think there's a process to go through.

terminator356:
Otherwise... wow, I dunno, probably more important to keep the text translations
 so close to a final release. Possible reversion required here?
Anyone with thoughts on what to do next?

kybos:
As to the translation issues:
1. I had started MusE in a couple of the localized languages prior to my changes. There were already many untranslated texts before (=displayed in English), so my recent changes did not cause the problem, they perhaps made it more visible.
2. Technically it looks like the locale/*.ts files have not been updated for a long time, so it would not even have been possible for translators to update the translations. 
3. There is a cmake build flag UPDATE_TRANSLATIONS for that (it's deactivated as a default). This is used to extract new/modified texts from the code and put them into the *.ts files. It also creates the runtime files (.qm) from these.
4. It seems that the Qt basic translation functionality (using the tr() macro) is not able to keep the mapping for modified texts, so they also have to be re-translated. (Qt obviously supports another, ID-based concept for translation that is able to do that, but I haven't checked that in detail.)

I would assume that translation in general is a part of the release cycle and must always be updated before a new release. But I of course know nothing about the availability of MusE translators and how it's normally done here. I could possibly help with at least 2 of the languages if absolutely necessary.

You mention we are "close to final release". Is there a roadmap? When is it supposed to be?

terminator356:
Oh dear. I looked yesterday and saw that stuff had not been updated in a long time.
When users submitted new or updated .ts files I simply committed them.
Now I'm not sure if those translations even made it into the program.
Maybe I was supposed to execute those steps.

Also, in the locale cmake file I neglected to add the 'components' folder,
 which contains stuff that was separated out of the 'widgets' folder.
I will do that, and make sure the list of folders in that cmake file is correct,
 and will check top level cmake(s) for trouble.

Bit of a shame. Been neglected for a while. What a mess.

In terms of sparse available translators (we don't really know who's out there),
 the release date would be the equivalent of...
 tomorrow ;-)

terminator356:
> Qt obviously supports another, ID-based concept
Ah, no wonder Orcan kept adding the more complex translation statements
 (QT_TRANSLATE_NOOP and so on).
I was always scared by them, not sure how to use them, so I've always just used tr().
I guess that was a mistake that has come back to bite.

spamatica:
> Maybe I was supposed to execute those steps.
I too had forgotten all about the extra step to create the source files to
base translations on but it's listed in the README.translate. If I recall
correctly the steps to update the .ts file are done to prepare for a new
translation so it is done by the translator before beginning the process.
If I read it correctly, adding the .ts to the source tree should be enough
to provide translations.

I did the Swedish translation long ago, sometimes I ponder removing it as
it nearly causes more issues than it fixes..  It is in dire need of being
updated and it is really hard work to find proper expressions for many
texts, very easy to end up with a valid sentence but in context it doesn't
really make sense.

kybos:
Yes, the README.translate contains the same instructions as the wiki page I mentioned above. 
If the translator does exactly as described there, then it should indeed be sufficient to commit the updated .ts files. The runtime .qm language files are created automatically at build time, independent of the UPDATE_TRANSLATIONS flag.

I still think that the consistency of the English texts should have the highest priority, as the English version is used by most and also to have a clean basis for translation into other languages. If it's OK with you, I would correct a couple more wrongly capitalized texts I've discovered in the meantime (really only a few, not more than 10 ;-) ). Then the potential translators could be notified, if feasible.

spamatica:
Den tis 3 dec. 2019 kl 18:04 skrev kybos <notifications@github.com>:

> I still think that the consistency of the English texts should have the
> highest priority, as the English version is used by most and also to have a
> clean basis for translation into other languages. If it's OK with you, I
> would correct a couple more wrongly capitalized texts I've discovered in
> the meantime (really only a few, not more than 10 ;-) ). Then the potential
> translators could be notified, if feasible.
>
I'd say just go ahead and fix :)

terminator356:
Agreed.
It might be possible to salvage some of what we have, by cross-checking
 in Linguist what the original texts were, and trying to re-map them.
Although, that might be a lot of work! I'd be happy if we could at least salvage
 whatever menu texts we have. I will look into it to see what I might do.

terminator356:
I suppose putting a call out to translators now, or close to release
 when we're certain of texts, is best rather than at some earlier alpha phase.
So it is good that we're having this conversation now.

kybos:
The Linguist apparently remembers and displays previous translations (or similar ones from other entities) and offers them for reuse. So they are just a click away but must be selected and confirmed manually.

PR #712, opened 2019-12-03
Text fixes

kybos:
Refer to #711.
Got a little more than intended, I used regex to search for them... But this should be it now.

There is a problem with the "Empty" text in the rack. It is not marked with tr() so it's not translatable, which is not good, as it's quite visible.
However, it can't be translated as it is now, because it would break the functionality - the text is used for direct string comparison:

`if (pipe->name(idx) == QString("Empty")) `
(rack.cpp, line 475)

It's being compared with:

```
QString Pipeline::name(int idx) const
      {
      PluginI* p = (*this)[idx];
      if (p)
            return p->name();
      return QString("Empty");
      }
```
(plugin.cpp, line 1463)

### DIFF
--- a/muse3/muse/app.cpp
+++ b/muse3/muse/app.cpp
@@ -2960,7 +2960,7 @@ bool MusE::checkRegionNotNull()
       if (end - start <= 0) {
             QMessageBox::critical(this,
                tr("MusE: Bounce"),
-               tr("set left/right marker for bounce range")
+               tr("Set left/right marker for bounce range")
                );
             return true;
             }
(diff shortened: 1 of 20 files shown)


terminator356:
Thanks. I'll merge now, and check that problem to see if we can eliminate that.

terminator356:
Eliminated all but one tr("Empty") string usage in plugins/pipeline/rack, at plugin.cpp : 1463.
A few other checks elsewhere for pipeline null.
Tested OK so far.

Pardon my work in progress on Jack transport master issues, in git master.
A new icon on the transport.
I finally found a way to ask absolutely whether we are transport master.
And now I think there may be a simpler more elegant way...

terminator356:
Forgot to mention: I have the Python cmake option enabled.
Some of that module may have things relevant to what we're fixing.
Advisable to enable it for build tests.

PR #713, opened 2019-12-04
Few more text fixes

kybos:
Fixed "db" -> "dB" and removed tr() for it, I think "dB" should not be translatable (as unit of measure).
Updated obsolete reference to MusE 2.0 in README.translate (corresponding Wiki page should be updated too?).
Fixed some more menu texts.

I noticed that File->New also triggers a dialog (template selection), so changed to New... here too.
An idea: Two entries, "New" to load default template (as when freshly started), and "New from Template..." for template selection (like e.g. in Inkscape and others).

### DIFF
--- a/muse3/README.translate
+++ b/muse3/README.translate
@@ -37,7 +37,7 @@ Step 4:
      Save the edited file "muse_fr.ts" from linguist and
      start File->Release. This generates the file "muse_fr.qm".
      Copy this file into your muse installation folder,
-     <prefix>/share/muse-2.0/locale/
+     <prefix>/share/muse-3.1/locale/
 
       or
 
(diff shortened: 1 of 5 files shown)


terminator356:
Thanks. No objections here to the idea.

kybos:
A thought on the "Window Config" top menu:
I already changed that to WinConfig to avoid the two words, it's better but still not optimal. Non-standardized abbreviations are generally not recommended in UIs, and something like that is very difficult to translate into other languages. 
I looked at the entries of this menu (in different views), and after some thought I think the best fitting English single word reflecting the content sufficiently would be "Display", and it's also easily translatable.
Any opinions on that?

terminator356:
Hm... Display... Lemme think about that. Has a nice ring to it though...
Just popping in to explain something in case you didn't see it.
If you're wondering why those items weren't moved to the Windows menu or elsewhere,
 look closely, Florian added this whole 'sharable menus and toolbars' system,
 which allows quite compact 'Mac-like' or 'Borland-like' UI behaviour (see global settings).
So, some of those menu items are sharable and can't be moved into other main Arranger menus.

kybos:
Yes, I already noticed that, so I was not suggesting getting rid of it completely, we just need a better text IMO...

terminator356:
Display. No objections here. I'd say go ahead.

PR #714, opened 2019-12-04
Tiny font in mixer strip tooltips

kybos:
I found out at last how to fix the tiny font problem in mixer tooltips, it did not seem possible...

This code line created an invalid style sheet (acc. to Qt docu and CSS specs), without any selector specified:
`setStyleSheet(MusECore::font2StyleSheet(MusEGlobal::config.fonts[1]));`

Funnily it still worked but it was not possible to change the style for anything else after that.
With the fix the tooltips will have the standard size.

### DIFF
--- a/muse3/muse/mixer/astrip.cpp
+++ b/muse3/muse/mixer/astrip.cpp
@@ -1351,8 +1351,9 @@ AudioStrip::AudioStrip(QWidget* parent, MusECore::AudioTrack* at, bool hasHandle
       
       // Set the whole strip's font, except for the label.
       // May be good to keep this. In the midi strip without it the upper rack is too tall at first. So avoid trouble.
-      setFont(MusEGlobal::config.fonts[1]);  
-      setStyleSheet(MusECore::font2StyleSheet(MusEGlobal::config.fonts[1]));
+//      setFont(MusEGlobal::config.fonts[1]); // should be redundant, overridden by style sheet
+      setStyleSheet("QWidget {" + MusECore::font2StyleSheet(MusEGlobal::config.fonts[1]) + "}" +
+                    "QToolTip {font-size:" + QString::number(MusEGlobal::config.fonts[0].pointSize()) + "pt}");
 
       channel       = at->channels();
 


terminator356:
Good catch. Thanks.
Don't forget that the same code is used in AudioStrip::configChanged() : 876
Also, the same two codes are used in MidiStrip ctor and  MidiStrip::configChanged().
The configChanged() happens (among many other things) when someone
 changes a font in settings.

You'll see I have a line in the configChanged()'s :
`if(font() != MusEGlobal::config.fonts[1])'
That is why I have those setFont() in there as well as the setting of stylesheet,
 to cover all bases, and that we have something to compare to.
If I recall there were certain scenarios where I needed the setFont(),
 like this comparison scenario.
However, the comparison may simply be an over-cautious redundancy.
You may remove it if we can be guaranteed that all will work OK,
 that the fonts are guaranteed to be what they're supposed to be,
 and there aren't too many serial redundant font changes.
That's unlikely since configChanged() is not usually called many times serially in a row.

I am a bit puzzled.
I use font2Stylesheet() 16 times in the app. They did not require putting the 
 QWidget {} wrapper around it as you have here.
Ah wait, sorry, I see. It is required to differentiate between the widget and the tooltip.

Before I merge this, would you like to review the other usages of font2Stylesheet(),
 and as mentioned fix up MidiStrip in mstrip.cpp ?
Thanks.

kybos:
Sure, I will look into it tomorrow.

terminator356:
Merged and we go from here. Looking at usages now...

kybos:
> I use font2Stylesheet() 16 times in the app. They did not require putting the
> QWidget {} wrapper around it as you have here.

I would be careful about that. The specs (Qt and CSS) always require a selector, so this is obviously a kind of "undocumented feature", and as such could be removed any time (in theory).

Qt documentation:
"Style sheets consist of a sequence of style rules. A style rule is made up of a selector and a declaration. The selector specifies which widgets are affected by the rule; the declaration specifies which properties should be set on the widget." ([https://doc.qt.io/qt-5/stylesheet-syntax.html](url))

To make sure that the behaviour remains exactly the same, I will add the generic selector * {...} to all usages. Thought about integrating it in the function itself, but that's not possible because of usage in string concatenation like that:
`stxt += MusECore::font2StyleSheet(fnt);`

terminator356:
Maybe if using setStyleSheet() it assumes that '*' selector or maybe QWidget is meant.
It just works.
I think I got that and saw it in setStyleSheet() examples. I think it's allowed.
But yeah, I can see now it's important to differentiate which font we want.
I will keep an eye open for that.

PR #722, opened 2019-12-05
Tooltip size fix

kybos:
Refers to #714.
I created a central wrapper function to ensure the tooltips always keep the same size, I think it's cleaner than to change each usage manually.
IMO changing tooltip font size looks like a mistake to the user, and there even is a bug in Qt related to that:  The tooltip frame is not always drawn correctly when the tooltip font size changes.

### DIFF
--- a/muse3/muse/components/utils.cpp
+++ b/muse3/muse/components/utils.cpp
@@ -43,6 +43,7 @@
 #include "part.h"
 #include "utils.h"
 #include "xml.h"
+#include "gconfig.h"
 
 namespace MusECore {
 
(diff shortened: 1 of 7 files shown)


terminator356:
Thanks. I had already started some fixes, just had some small conflicts.
Hang in there I'll change my few remaining lines to use your new wrapper.

I am having (long time) propagation problems, trying to solve.
Font changes not propagated from mixer midi strip down to component rack, or down
 to the 'patch edit' controls inside component rack. Fixed the former but latter still not working.
New snooper fix shows **ALL** object properties, including inherited ones such as font.
Revealed that fonts are not being propagated to these controls.

terminator356:
OK my lines fixed. Thanks. Proceed.
My propagation problem still there :-(

PR #723, opened 2019-12-05
Added File->New from template

kybos:
As discussed in #713.

I get the message "Can’t automatically merge. Don’t worry, you can still create the pull request. ". 
Is that a problem? Should I merge/rebase with current master first?


kybos:
OK, it's only the change log, should be easy to resolve.
I am not able to do it, no write access to repo.

terminator356:
Fixed. Thanks.

kybos:
Thanks.
Not sure if "New from template" action should be added to the toolbar as well?

terminator356:
Sounds good. Go for it!
(FYI If ever you need to create a new toolbar, heed the NOTICEs in there about toolbar names.)

PR #724, opened 2019-12-05
Replace WinConfig by Display

kybos:
Refer to #713.

### DIFF
--- a/muse3/muse/arranger/arrangerview.cpp
+++ b/muse3/muse/arranger/arrangerview.cpp
@@ -261,7 +261,7 @@ ArrangerView::ArrangerView(QWidget* parent)
   functions_menu->addAction(editCleanPartsAction);
   
   
-  QMenu* menuSettings = menuBar()->addMenu(tr("Win&Config"));
+  QMenu* menuSettings = menuBar()->addMenu(tr("&Display"));
   menuSettings->addAction(tr("Configure &Custom Columns..."), this, SLOT(configCustomColumns()));
   menuSettings->addSeparator();
   menuSettings->addAction(subwinAction);
--- a/muse3/muse/cliplist/cliplist.cpp
+++ b/muse3/muse/cliplist/cliplist.cpp
@@ -133,7 +133,7 @@ ClipListEdit::ClipListEdit(QWidget* parent)
       setCentralWidget(editor);
 
             
-      QMenu* settingsMenu = menuBar()->addMenu(tr("Win&Config"));
+      QMenu* settingsMenu = menuBar()->addMenu(tr("&Display"));
       settingsMenu->addAction(subwinAction);      
       settingsMenu->addAction(shareAction);      
       settingsMenu->addAction(fullscreenAction);      
--- a/muse3/muse/liste/listedit.cpp
+++ b/muse3/muse/liste/listedit.cpp
@@ -513,7 +513,7 @@ ListEdit::ListEdit(MusECore::PartList* pl, QWidget* parent, const char* name)
       connect(incAction,    &QAction::triggered, [this]() { cmd(CMD_INC); } );
       connect(decAction,    &QAction::triggered, [this]() { cmd(CMD_DEC); } );
 
-      QMenu* settingsMenu = menuBar()->addMenu(tr("Win&Config"));
+      QMenu* settingsMenu = menuBar()->addMenu(tr("&Display"));
       settingsMenu->addAction(subwinAction);
       settingsMenu->addAction(shareAction);
       settingsMenu->addAction(fullscreenAction);
--- a/muse3/muse/marker/markerview.cpp
+++ b/muse3/muse/marker/markerview.cpp
@@ -190,7 +190,7 @@ MarkerView::MarkerView(QWidget* parent)
       editMenu->addAction(markerDelete);
       
       
-      QMenu* settingsMenu = menuBar()->addMenu(tr("Win&Config"));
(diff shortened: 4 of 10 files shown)


kybos:
Noticed one thing: 
The menu entry "Fullscreen" resides both in WinConfig (now Display) and View menus, but only the latter one is active. Is that intentional?

terminator356:
Yes. The one under Display becomes active when the window is outside the MDI.

terminator356:
I'm thinking of moving the items under the Windows menu to the View menu.
Any caveats? Objections?

spamatica:
No objections.

To speak freely about related features, the different view styles we
"support", mdi, sdi, borland etc. I wonder if anyone actually uses anything
but standard? I guess the variants work but to what extent, no idea. I do
know they add a huge amount of test-scenarios.

Another thing, the Appearance dialog should really be moved to the Global
Settings dialog. There's a bug report for it somewhere.

Den tors 5 dec. 2019 kl 22:18 skrev Tim <notifications@github.com>:

terminator356:
I think I had a comment recently that someone (emphasis one?) uses it.

I think it is certainly useful, as Florian said "I think MusE wastes a lot of space."
He did have a point there. All those editor windows.
I have a soft spot for the Borland style - I coded with their tools for many years,
 and actually suggested that theme!

Once you get into making a big song you can probably dig it ;-)
And... you're looking at it from a 2020 hidpi point of view, it was added say, 8 years ago?

Code wise... Yeesh, well, yeah I know. It's a big overhead. A lot to keep an eye on.
Today I just noticed that when reattaching the windows the menus may screw up with
 items on top of each other. I've been maintaining that stuff and tested it fine a while ago,
 if I get time I'll check, hopefully some neglected maneuver... (my luck no.)

I propose a poll on the forum.
Then again... removing it would be nightmarish.

FYI There were several strong requests for docking features, here somewhere I think.
I was thinking today about how that would mesh with these features.
Docking instead of pure MDI? Or along side it? Not sure.

Finally, I've always wondered:
What actually happens with that Mac theme, if we were able to run MusE on a real Mac?

PS: You're invited. Hurry, c'mon in. I want to try the 'team' discussion features
 to see if it's better than using this way! ;-)

terminator356:
Appearance and others: Yes, it would be good to merge some of those settings dialogs
 into our Settings dialog with more tabs. It's been on my mind lately.
We must be careful. Some of the dialogs are designed to stay open as 'companions'
 if desired, as the user works.
Appearance has the 'live' colour editor.
It uses a 'backup' and 'working' copy of appearance settings.
Still if we can result in a final settings dialog being able to be resized to a
 'companion' size, then there's no difference other than one extra click to get to the tab.

kybos:
As for Appearance, just my opinion:
I don't think it's bad or wrong to have a separate menu for that, this is not unusual. IMO better than one setting dialog with too many tabs (in that case a tree-based approach would probably be more ergonomic, like in most large apps).
Perhaps:
- Put Appearance directly below Global settings.
- Change "Global settings" to "Preferences" (more standard, or alternatively Options, or just Global).
- Change "Appearance settings" to "Appearance" (the whole menu is called Settings, so it's redundant).

kybos:
> To speak freely about related features, the different view styles we "support", mdi, sdi, borland etc. I wonder if anyone actually uses anything but standard?

I tend to use the Cakewalk-style MDI, but it has several issues - it always opens the sub-windows in windowed-mode (not maximized, even if I maximized it before), the window does not subject to the theming - the titlebar is always blue and the title text blurred.

spamatica:
> Then again... removing it would be nightmarish.

Yeah... I guess it's me being old and grumpy, removing it is probably not
an option ;)

> As for Appearance, just my opinion: I don't think it's bad or wrong to
have a separate menu for that

The problem I have with this is that it is hard to draw the line as to what
belongs where. There are many items in global settings, not the least GUI
style, which could as well be in Appearance. While looking for settings I
open the wrong dialog all the time.

> I tend to use the Cakewalk-style MDI
Interesting, one user at least then :) Though with varying degrees of
success...

spamatica:
kybos, I realize we(me atleast) don't seem to have a way to reach you
outside of the tickets :) Would you mind sending your mail address to me (
spamatica@gmail.com)?



Den fre 6 dec. 2019 kl 12:29 skrev Robert Jonsson <spamatica@gmail.com>:

> > Then again... removing it would be nightmarish.
>
> Yeah... I guess it's me being old and grumpy, removing it is probably not
> an option ;)
>
> > As for Appearance, just my opinion: I don't think it's bad or wrong to
> have a separate menu for that
>
> The problem I have with this is that it is hard to draw the line as to
> what belongs where. There are many items in global settings, not the least
> GUI style, which could as well be in Appearance. While looking for settings
> I open the wrong dialog all the time.
>
> > I tend to use the Cakewalk-style MDI
> Interesting, one user at least then :) Though with varying degrees of
> success...
>
>

kybos:
> I'm thinking of moving the items under the Windows menu to the View menu.
> Any caveats? Objections?

Perhaps one caveat: If you have many MusE windows open, then the Window menu itself can get pretty long. I wonder if appending all these to the (already relatively long) View menu wouldn't be too much?

terminator356:
Yeah I know. Worries me too. We'll see if it can be done nicely...

kybos:
> The problem I have with this is that it is hard to draw the line as to what belongs where. There are many items in global settings, not the least GUI style, which could as well be in Appearance.

True.
I guess the "Global" tabs View, GUI Behaviour and GUI Style would probably be better in Appearance?

PR #727, opened 2019-12-08
Option for opening subwindows maximized

kybos:
I was missing that for a sensible use of the Cakewalk-style MDI mode...


kybos:
Sorry, again a conflict in the change log.
Still not able to resolve it myself.

terminator356:
There appears to be a **BUG**(?) in GitHub's "Resolve Conflicts" feature.
I used it OK to fix some code in a cpp file and the 'mark as resolved' button became **enabled**.
But twice now (once with the previous conflict) it is **refusing** to do the same
 for the "ChangeLog" file. 
This is not good. I can't fix the conflict from here !
I can try the 'command line' method.
But first can you try anything on your side?
How about a direct push, since you're a member?
My disk space is really low and I have tons of thongs to work on ATM.

terminator356:
Ha! There was an extra "<<<< master" hiding on the page way down from several months ago!
Thanks. We may proceed gracefully now.

PR #728, opened 2019-12-08
Some more icon and text fixes

kybos:
New icons for event list editor
Fixed text capitalization in Settings dialogs


terminator356:
Thanks.

PR #729, opened 2019-12-09
Update translation files, translate German

kybos:
I updated the translation files for all languages (some of the texts are translated automatically - like text duplicates etc.).
I then translated the German file, I would say 80-90% of the texts are there now, really lot of work...
Some texts I don't quite understand, some are impossible to translate without context (like the single characters), some would sound very strange in German (like the Latency tab in the settings).

For cases like tr("I") the optional comment for the tr() macro can be used be developers to give the translator a hint, e.g. 

//: means Input
...tr("I");


terminator356:
Thanks. I... guess I'll just have to trust one this one, eh? ;-)

Those single-letter items must be a real pain. How to translate...
What would happen if we used the letters R(ecord) M(ute) and S(olo) on our buttons
 instead of icons, like so may DAWs? Is the concept translatable to say, German?
Oh wait - we use letters in the track header!
One time I argued with someone about that. Letters vs. icons.
Here, we are using both becuase of that header.
I thought about replacing the letters with the corresponding icons.
Seemed like a weird but maybe plausible idea...

kybos:
I think it's OK to keep the letters for the track header, for the user there are the tooltips, and for the translator I will add hints to the code (like I wrote above, you can put a special comment //: above the tr() line, then it will be displayed to the translator in Qt Linguist).

There are things like the following, which are practically impossible to translate, probably they should not be made translatable at all:

```
      columnnames << tr("Port")
		  << tr("Device Name")
		  << tr("s")
		  << tr("c")
		  << tr("k")
		  << tr("r")
		  << tr("m")
		  << tr("t")
		  << tr("type")
		  << tr("rid") // Receive
		  << tr("rc") // Receive
		  << tr("rr") // Receive
		  << tr("rm") // Receive
		  << tr("rt") // Receive
		  << tr("rw") // Receive
		  << tr("tid") // Transmit
		  << tr("tc") // Transmit
		  << tr("tr") // Transmit
		  << tr("tm") // Transmit
		  << tr("tt"); // Transmit
```

terminator356:
> (like I wrote above, you can put a special comment //: above the tr() line, then it will be displayed to the translator in Qt Linguist).

Oh I see, wasn't sure what you meant there. That's a neat comment trick for sure.
Hope I can remember it.

terminator356:
Go ahead and remove the tr() on those items if you like.
Doubtful they'd ever be translatable.

PR #730, opened 2019-12-09
Translation fixes

kybos:
Fixes for translation, as discussed in #729

### DIFF
--- a/muse3/muse/arranger/arranger.cpp
+++ b/muse3/muse/arranger/arranger.cpp
@@ -404,16 +404,23 @@ Arranger::Arranger(ArrangerView* parent, const char* name)
       QFontMetrics fm1 = header->fontMetrics();
       int fw = 11;
 
-      header->setColumnLabel(tr("#"), COL_TRACK_IDX, 10);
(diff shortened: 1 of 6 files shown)


terminator356:
Thanks for the great help on the translation stuff.
Hey, I mentioned I neglected to add 'components' to share/locale/CMakeLists.txt
Can you add it and see what happens? And any other folders that should be there?
I'm in another branch ATM...

kybos:
OK, will try that out tomorrow...
So there are still more texts coming ;-).

kybos:
> Hey, I mentioned I neglected to add 'components' to share/locale/CMakeLists.txt
> Can you add it and see what happens? And any other folders that should be there?

I added "components" and updated but there is no difference in the number of source texts.

terminator356:
Hm, funny I could swear there's a bunch 'o translatable text in those components. Maybe not?

kybos:
Yes there is, but I have the impression these texts were already there before.

terminator356:
I might have broken any links by moving many files from the widget folder
 to the components folder

PR #732, opened 2019-12-10
Track header icons

kybos:


terminator356:
Robert any thoughts? Are you OK with replacing the track header letters with icons?

terminator356:
Most of our custom headers have a context menu where you choose which columns to show.
In header.cpp : Header::mousePressEvent(), you'll need to put icons for some of the items.

How to? I suppose in that routine you can check if the item has an icon,
 and if so, show it instead of the letter that is used to compose the popup text.
I think that'll work... hopefully.

terminator356:
I'll merge this now since it is easily switchable (thanks for the comments).

kybos:
It is possible to create the context menu actions using the proper icon, but then this icon will be displayed instead of the checkbox at the beginning of the line (can't make a screenshot when menu is open). Is that better? There seems to be no way to put the icon instead of the text.
The issue is of purely cosmetic nature anyway, the functionality is not impaired by it, as far as I can judge, since there is always the description (the tooltip text) displayed. 

I created a PR with the icons in the context menu: #733. Please check and tell me what you think.

Alternatively only the descriptions could be shown in the context menu. The first part of the text mostly has not much added value anyway.

PR #733, opened 2019-12-11
Add header icons to context menu

kybos:
Refers to #732.

### DIFF
--- a/muse3/muse/components/header.cpp
+++ b/muse3/muse/components/header.cpp
@@ -158,13 +158,18 @@ void Header::mousePressEvent ( QMouseEvent * e )
     QAction* act = 0;
 
     for(int i=0; i < count(); i++) {
-      act = p->addAction(itemModel->horizontalHeaderItem(logicalIndex(i))->text() +
-                         "\t - "+ itemModel->horizontalHeaderItem(logicalIndex(i))->toolTip());
+        QIcon icon = itemModel->horizontalHeaderItem(logicalIndex(i))->icon();
+        if (!icon.isNull()) {
+            act = p->addAction(icon, "\t - "+ itemModel->horizontalHeaderItem(logicalIndex(i))->toolTip());
+        } else {
+            act = p->addAction(itemModel->horizontalHeaderItem(logicalIndex(i))->text() +
+                               "\t - "+ itemModel->horizontalHeaderItem(logicalIndex(i))->toolTip());
+        }
 
-      act->setCheckable(true);
-      act->setChecked(!isSectionHidden(logicalIndex(i)));
-      int data = logicalIndex(i);
-      act->setData(data);
+        act->setCheckable(true);
+        act->setChecked(!isSectionHidden(logicalIndex(i)));
+        int data = logicalIndex(i);
+        act->setData(data);
     }
     connect(p, SIGNAL(triggered(QAction*)), SLOT(changeColumns(QAction*)));
     p->exec(QCursor::pos());


terminator356:
Can't do much more than that without making custom menu items.
Accepted. Thanks.

terminator356:
> It is possible to create the context menu actions using the proper icon, but then this icon will be displayed instead of the checkbox at the beginning of the line

Ha! I can tell that you do all your MusE coding with the Ardour stylesheet active.
Here it is without a stylesheet:
Here we see a case where 'innocent' stylesheets can really do a number on your plans.
track header 1

kybos:
Not really, it always looks like that, even with the default style:
muse1
Probably styles in KDE vs. Gnome & al.?
In any case it shows that using tabulators in this way can have unpredictable results...

terminator356:
Wow. Bizarre! Yeah, must be the system styles wreaking havoc.

terminator356:
You are right. I knew that I had to be wrong about stylesheets being the cause.
It is the chosen style. With no stylesheet, when I select different styles it changes
 the behaviour.
And, I forgot about this, but our themes (wisely) force a style selection.
At the top of Ardour.cfg for example there is 
"theme Oxygen"
So that explains what's going on here.

terminator356:
It is for these types of reasons that, as I mentioned, the OOM fork of MusE
 completely discarded all programmable colours and styles,
 in favour of one big master stylesheet. Nothing left to 'unpredictability'.
I remember a quote from the lead dev: "MusE is not a desktop application".
I suppose he had a point. It is rather specialized in how it paints. Tolerances need
 to be very tight in some areas, it meshes badly with system if we leave it
 up to the system to paint.

You should have seen MusE in Qt2 or 3 before all of this stuff.
Man, edges and boxes and margins and controls were so tight you couldn't fit a needle through.
We could put a fixed height and knew it would stay that way for everyone.
Sort of like what you see with the current Qt Windows style.

Relevant: I was brainstorming about how to expose our configuration colours
 as properties so the stylesheet can access them directly without a separate .cfg file.
We'll see if I can do something. Tricky as usual.

terminator356:
Even with a complete custom stylesheet, as I understand it they cannot
 tell completely exactly how to draw a control. Only borders and padding and so on.
They are still somewhat at the mercy of whatever the current style is.
As we can see, it does influence the drawing. And forces us to force the style.

kybos:
> At the top of Ardour.cfg for example there is
> "theme Oxygen"

It seems that in the other style-cfg.s there is no predefined theme.
I am wondering what effect Oxygen has on a non-KDE desktop - I have no Oxygen style elements installed.

terminator356:
Maybe it just defaults to your current system setting.

I just tried the 'Adwaita' style. There are no icons at all in any of the menus.
Phht! Sometimes I hate dealing with all of these different styles.
The hours I've spent trying to make everything look right with all of them.

kybos:
I also noticed another weird thing:
In Appearance, when I change any setting (not necessarily a font-related one, or even none) and press Apply or OK, the font in MusE is always changed. Weird as I never messed up with the font settings in MusE or in my Linux system...

terminator356:
Hey want a good laugh? Here's 'bb10' style:
Wee! Looky them drop down combo boxes! I'll stick to 'official' styles thank very much ;-)
ugly style 1

terminator356:
No, that big circle bottom left is NOT a volume control, he he.
It is the actual knob on the h-slider ! Wee !

PR #735, opened 2019-12-11
Configurable size for icons and custom cursors

kybos:
Make icons and custom cursors size configurable


terminator356:
Hey, now I can make them as big as cell phone icons!

PR #738, opened 2019-12-13
Fix icons from theme

kybos:
I hope it's OK now.

I was mainly trying to avoid all those conversions and additional pixmaps for my new icons, because the Qt icon engine can handle SVG directly since 4.2. Possibly a cleanup would be good also for the remaining SVG icons to avoid the overhead.


terminator356:
OK Thanks very much. This looks like what I was thinking.
We'll see what shakes out...

PR #739, opened 2019-12-14
Cleanup obsolete version checks, minor text/icon changes

kybos:
MusE requires Qt 5.7 as a prerequisite, so removing the obsolete version checks.


terminator356:
Good stuff. Thanks.

PR #741, opened 2019-12-16
Style rack scrollbar

kybos:
As discussed in dev space...


terminator356:
Thanks very much for that!

PR #749, opened 2020-01-09
Fix track list header for better adaptability

kybos:
Should fix some sizing issues, especially with the track list header, to better adapt to different font sizes and longer translations.


terminator356:
Welcome back. Thanks for spotting the RubberBand fix, my bad, I'll fix it up.

PR #750, opened 2020-01-11
Appearance dlg - fixes for styles and font0

kybos:
Some modifications from last year that I did not want to merge before my vacation, in case there were issues...

Appearance dialog:
1. Removed the inconsistent handling (special button) for changing the MusE color scheme. Now the scheme is switched by Apply/OK like all the other settings.
2. Implemented user override for the internal colour settings. Now if there is a file with the same name as the scheme + .cfc (like Ardour.cfc) in the user theme directory, then it will be applied at scheme change.
3. Looking into the font handling in more depth, I think Font 0 can't be consistently applied and is more or less redundant. It's always replaced by the system font (Qt theme font) and/or style sheet font at MusE start. 
Implicitly changing it when appearance settings are applied only adds to the confusion, because the behaviour is different depending on the desktop used (and probably never intended by the user), and is lost at the next MusE start anyway.
So I changed all the usages of font0 (mostly my early fixes of too small fonts) to the current application font instead. Font 0 is still used in main.cpp as base setting before style/stylesheet is applied - just to be sure, in case there is a scenario where it's necessary. Even so it could be replaced by a constant there and removed from the appearance settings IMO to avoid confusion.

### DIFF
--- a/muse3/muse/app.cpp
+++ b/muse3/muse/app.cpp
@@ -315,39 +315,38 @@ void addProject(const QString& name)
 MusE::MusE() : QMainWindow()
       {
       setIconSize(QSize(MusEGlobal::config.iconSize, MusEGlobal::config.iconSize));
-//      setIconSize(ICON_SIZE);
       setFocusPolicy(Qt::NoFocus);
       MusEGlobal::muse      = this;    // hack
       _isRestartingApp      = false;
-      clipListEdit          = 0;
-      midiSyncConfig        = 0;
-      midiRemoteConfig      = 0;
-      midiPortConfig        = 0;
-      metronomeConfig       = 0;
-      midiFileConfig        = 0;
-      midiFilterConfig      = 0;
-      midiInputTransform    = 0;
-      midiRhythmGenerator   = 0;
-      globalSettingsConfig  = 0;
-      markerView            = 0;
-      arrangerView          = 0;
-      softSynthesizerConfig = 0;
-      midiTransformerDialog = 0;
-      shortcutConfig        = 0;
-      appearance            = 0;
-      _snooperDialog        = 0;
+      clipListEdit          = nullptr;
+      midiSyncConfig        = nullptr;
+      midiRemoteConfig      = nullptr;
+      midiPortConfig        = nullptr;
+      metronomeConfig       = nullptr;
+      midiFileConfig        = nullptr;
+      midiFilterConfig      = nullptr;
+      midiInputTransform    = nullptr;
+      midiRhythmGenerator   = nullptr;
+      globalSettingsConfig  = nullptr;
+      markerView            = nullptr;
+      arrangerView          = nullptr;
+      softSynthesizerConfig = nullptr;
+      midiTransformerDialog = nullptr;
+      shortcutConfig        = nullptr;
+      appearance            = nullptr;
+      _snooperDialog        = nullptr;
       //audioMixer            = 0;
-      mixer1                = 0;
-      mixer2                = 0;
-      routeDialog           = 0;
+      mixer1                = nullptr;
+      mixer2                = nullptr;
+      routeDialog           = nullptr;
       watchdogThread        = 0;
-      editInstrument        = 0;
+      editInstrument        = nullptr;
       //routingPopupMenu      = 0;
-      progress              = 0;
+      progress              = nullptr;
       saveIncrement         = 0;
-      activeTopWin          = NULL;
-      currentMenuSharingTopwin = NULL;
-      waitingForTopwin      = NULL;
+      activeTopWin          = nullptr;
+      currentMenuSharingTopwin = nullptr;
+      waitingForTopwin      = nullptr;
 
       appName               = PACKAGE_NAME;
       setWindowTitle(appName);
@@ -1029,7 +1028,7 @@ MusE::MusE() : QMainWindow()
       }
 
       transport = new MusEGui::Transport(this, "transport");
-      bigtime   = 0;
+      bigtime   = nullptr;
 
       MusEGlobal::song->blockSignals(false);
 
--- a/muse3/muse/components/appearance.cpp
+++ b/muse3/muse/components/appearance.cpp
@@ -303,7 +303,7 @@ Appearance::Appearance(Arranger* a, QWidget* parent)
       //---------------------------------------------------
       //    THEMES
       //---------------------------------------------------
-      connect(changeThemeButton, SIGNAL(clicked()), SLOT(changeTheme()));
+//      connect(changeThemeButton, SIGNAL(clicked()), SLOT(changeTheme()));
 
       QDir themeDir(MusEGlobal::museGlobalShare + QString("/themes"));
       QStringList list;
@@ -687,17 +687,17 @@ void Appearance::updateFonts()
 //
 //---------------------------------------------------------
 
-void Appearance::changeTheme()
+bool Appearance::changeTheme()
 {
     if (colorSchemeComboBox->currentIndex() == 0) {
-      return;
+      return false;
     }
     if(QMessageBox::question(MusEGlobal::muse, QString("Muse"),
       tr("Do you really want to reset colors to theme default?"),
       QMessageBox::Ok | QMessageBox::Cancel, QMessageBox::Ok)
        != QMessageBox::Ok)
     {
-        return;
+        return false;
     }
 
     backgroundTree->reset();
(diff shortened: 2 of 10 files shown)


terminator356:
Will there still be an overall "font 0" or similar where the user can quickly change overall main font?
If our so-called "font 0" is not retaining it's effect on restart then there's something wrong.
It should be working normally, but maybe with a stylesheet applied it gets overwritten?
I do understand about the appearance settings and stylesheet overwriting each other,
 a mess which I recognized and acknowledged before the holiday, but at least the
 main font should keep it's effect on restart I think.

kybos:
I would prefer that too (I was trying really hard...), but after a lot of testing with various desktops and scheme/style settings I am almost sure that it's not possible in a consistent way.
There is this function called in main.cpp:

`void updateThemeAndStyle(bool forceStyle)`
`{`
 ` // Note that setting a stylesheet completely takes over the font until blanked again.`
 ` qApp->setFont(MusEGlobal::config.fonts[0]);`
 ` loadStyleSheetFile(MusEGlobal::config.styleSheetFile);`
  `loadTheme(MusEGlobal::config.style, forceStyle || !MusEGlobal::config.styleSheetFile.isEmpty());`
`}`

So font0 is always overridden by something else, Qt style/theme and/or stylesheet. But even if the first two lines are commented out, font0 is not kept for some reason... I also tried to set the font0 for the MainWindow explicitly, but the results are inconsistent, depending on the desktop used and other factors.

It seems that the only way to override the font reliably is to set it in the style sheet. So you theoretically could read the external stylesheet at start (if available) and merge the font0 settings with it (for QWidget), but I would consider that quite an evil hack and bad "style". 

It looks like it is the intention of Qt to force the main font from system/Qt settings (and stylesheet, if used), so perhaps it's not so a bad thing to do so? Not sure, what do you think?

BTW: I did not remove font0 in this PR, only changed some references to use the actual main font instead. To remove font0 completely was only a suggestion, as I got the impression that it was in fact redundant.

terminator356:
[Sigh...] I see that with a stylesheet it is not retaining font 0 for much of the app
 (places we don't specifically set it), despite apparently no font settings
 in the stylesheet or otherwise. (I guess that's why I wrote that warning comment, eh?)
I may have also tried setting the font _after_ setting the stylesheet (or did I?)

OK Let's get these fixes in, and we'll figure something out. Thanks!
Help on QApplication::setFont() does mention setting the app font via stylesheet,
 so maybe we can simply use our nifty font2stylesheet functions.

Does "font 0" really affect me? Personally no. All these years I've tried to retain it
 simply because we could, and justify it by thinking about for example, the visually impaired
 so they could quickly tweak it to something higher.
But they would likely change a system setting for that, not just MusE !
There may be some who use that setting, I may have seen it in screen grabs.

Regardless of what we end up doing with the app font and some of the others,
 there is still a need for a few different font sizes throughout the app.
Especially the mixer strips where things need to be slightly more compact.
The strips look just a bit too big and ugly (intrusive?) when they're the same font as the app.

Proposal: If we remove font 0 and let it be dictated by the system or stylesheet font
 (which actually sounds reasonable the more I think of it), then **absolute** font sizes for
 the other fonts becomes **meaningless**. We could change them to mean **delta** values.
ie. User: "Gimme a strip font oh, a size or two **smaller than** whatever the app font is,
 so I can view more strips at once on the screen in my big project."

kybos:
> Regardless of what we end up doing with the app font and some of the others,
> there is still a need for a few different font sizes throughout the app.

Absolutely, no doubt about that.

Your idea with the relative font sizes sounds interesting. So we could consider main font size of 10 to be 100% and scale the other fonts accordingly if main font <> 10. I will look into it.

terminator356:
Mm, not exactly. The number entry boxes beside the fonts would need to become offsets
 like -2, -1, 0, +1, +2 etc. Actually you raise a point, that it ought to be factors instead,
 something like x0.1, x0.2, x0.3 etc. so that it proportions well with larger base fonts.
There would be no font 0. We would leave the base app font alone and don't touch it.
Then these 'offsets' or 'factors' apply to whatever the current base font happens to be.
For example if Font 1 says "x0.8" with a reported base app font size of 12, the
 resulting Font 1 size actually used internally would be 9.6 (say floored to 9).

kybos:
I looked a little and it seems quite a lot of work to implement. There are so many calls of setFont...
Not sure if it's really worth it? The user can easily change the absolute font sizes if he wants to. We could even add a button on the appearance font tab to dynamically insert values into the size fields in relation to the current main font, according to the fixed factors. Would be much easier and less intrusive...

I think important thing is to remove font0 as it simply does not work as expected and thus looks like a bug to the user. And text like Font 0 should not appear on a UI anyway, IMO, as normal people (non-programmers ;-) ) start counting with one (I also had usability in mind as I had suggested percentages for the relative font sizes, I think that is what people are accustomed to, more than the rather technical floating point factors). 

I would suggest to display the main font as a (non-changeable) reference in the place of font 0, like this:
image

Let me know what you think. If you still prefer the full-fledged solution with the relative font sizes, I will tackle it.

terminator356:
Percent: Yes good. Duh, what was I thinking with decimal points...
I agree with the picture regarding 'font 0'. Looks good to me, good to show what
 the current state is, especially if we are to show percentages in the other boxes.

> The user can easily change the absolute font sizes if he wants to. We could even add a button on the appearance font tab to dynamically insert values into the size fields in relation to the current main font, according to the fixed factors.

Not sure about that, if I understand correctly. So the boxes would still say 7, 10, 8...
 but internally they are scaled according to the main font? Seems like that might
 be confusing, user thinks it is '10' but it is not?
Hm... It might be possible to accomplish what we want very simply, **not** requiring
 changing all the places using these fonts. Here's how:
The actual boxes would show a percent. But our actual internal array of fonts
 would hold the real final fonts and their sizes. When a user adjusts a font percentage,
 the internal font is adjusted accordingly, and from there, nothing further need be done,
 all the places using the fonts 'just work'.
Something like that? Maybe that's what you also had in mind?
I think that might work OK!

kybos:
Yes, that's better than what I had in mind, thanks!

terminator356:
One further change of course would be the xml storage of the fonts in the config file,
 where we would store the percentage value instead of the absolute font size,
 but beyond that quite possibly not much more would need to be done, I think, hopefully.

kybos:
So if I see it right that would entail:
- Storing the percentages/factors additionally for each font in the configuration.
- Reading the current app font size at every program start (probably in the MainWindow constructor in case the font was set in the style sheet for QWidget)
- Calculating the font sizes using the main font and the stored factors, and saving the point values to the configuration.
- In the font settings, replace pt sizes by percentages and let the user change those if desired.

Is that more or less what you had in mind?

BTW, my original proposal was very simple (sorry for not being clear enough):
1. Just keep the absolute sizes for the non-main fonts, no automatic scaling, leave it to the user to set the sizes he likes, in Appearance or in the style sheet.
2. As a small help, offer a function (a button in the font settings in Appearance) to calculate and insert recommended sizes according to fixed ratios and the current main font size.

Perhaps not very sophisticated, but transparent and easy to understand for the user, IMO.

Not sure which option is the better one... Opinions?

kybos:
Sorry, did not see your last comment, but I think my reasonings are still more or less valid ;-).

terminator356:
Some further info about the xml storage, policies, and recommended action:
I see it's gonna be a bit more work. In conf.cpp we have these:

`                        else if (tag == "font1")
                              MusEGlobal::config.fonts[1].fromString(xml.parse1());
`
and 
`      for (int i = 0; i < NUM_FONTS; ++i) {`
`            xml.strTag(level, QString("font") + QString::number(i),` `MusEGlobal::config.fonts[i].toString());`
`            }`

So we'll have to come up with a way to store the percent instead of size, in that string :-(
Hopefully simply compose / decode the string separately and replace size with percent.

Regarding existing user settings, when altering their nature in the program:
There are three main situations usually encountered, in order of importance:
1) A new Qt header column is added somewhere for example.
This requires using our songfile versioning mechanism, bumping MusE's songfile version
 up, checking it against actual song version upon load, and of course saving the new version.
The existing binary Qt header column data in the song is incompatible with the new data
 and would result in corrupt header arrangement. So the old data is _ignored_ and fresh
 data stored. The songfile version mechanism is used to determine whether to do that or not.
 
2) A MusE setting's nature, name or type is changed but existing setting can still be imported.
Usually does not require bumping up songfile version. Simply quietly change the 
 nature of the setting upon saving.
Here we must (sometimes cleverly) change the tag name of the setting upon save
 to distinguish from the old one. So existing integer "samplerate" might become
 float "samplerateF" or something like that.

3) A Muse setting's nature, name or type is changed and existing setting cannot be imported.
Similar to number 2, but here an existing setting is awkward or makes no sense,
 or is completely eliminated. So we have no choice to ignore the setting and save
 a fresh new setting.

Usually, we try to 'import' existing settings when loading - 'shoehorning' the settings
 into however the newer code works, then upon save, the 'newer' settings and tag names
 (which may change) are saved. Essentially this discards the old settings after importing them.
In the case of 2 it can be done, in 3 it cannot.

So... I think in this case of fonts, we _can_ use number 2.
But we will ignore (somehow) the font 'size' part of the string upon loading.
Don't bother trying to 'import' the existing size. If you can great but no worries.
(Don't forget to give the tags new names.)
You can go with number 3 as well. Ignore all existing font settings and just start with new ones.
No one will complain if these non-critical settings are lost upon load. Makes your job easier.

How to 'ignore' an xml setting tag:
1) Read the setting into an appropriate (dummy) variable and simply discard it out of scope.
or
2) I think you can simply use parse1() like this:
`                              // Same as unknown() without the error message.`
`                              xml.parse1();`
I've used it, but I'm not sure if it is really the right way, it attempts to read the object
 but essentially ignores it so in the case of a corrupt file it might fail, but I guess that's that.

terminator356:
Ha, posted before your post - spaghetti replies ;-)

kybos:
Thanks for the information on the configuration storage.
I started to implement the relative font sizes, sure it can be done, but I am still not quite convinced that it's worth the effort, considering the config incompatibilities, new config options required etc. Also some usability issues: When the user selects a font in the standard font dialog, he expects to be able to set the size - we would have to ignore it implicitly (or calculate the percentage dynamically, which is not very transparent).
I think that my suggestion below (keeping the absolute sizes) would be more transparent and sufficient for the (probably very few) users that use a larger main font size.

Please let me know your opinion, I will leave the final decision to you ;-).

> BTW, my original proposal was very simple (sorry for not being clear enough):
> 
>     1. Just keep the absolute sizes for the non-main fonts, no automatic scaling, leave it to the user to set the sizes he likes, in Appearance or in the style sheet.
> 
>     2. As a small help, offer a function (a button in the font settings in Appearance) to calculate and insert recommended sizes according to fixed ratios and the current main font size.
> 
> 
> Perhaps not very sophisticated, but transparent and easy to understand for the user, IMO.
> 
> Not sure which option is the better one... Opinions?

terminator356:
> When the user selects a font in the standard font dialog...
Ouch. Hmm... Yeah bummer, eh. Shot down...?
I did read your proposal before. I think I had trouble visualizing it, not quite sure yet.
It's late. Lemme check your PR tomorrow and think some more.

kybos:
No hurry, take your time.
BTW: My other PR does not (yet) implement this suggestion, it just replaces font 0 by the read-only display of the actual system font, as in the screenshot above. So it should be safe to merge.

terminator356:
Merged PR. Your proposal is looking like the better (only) way...
Also in light of the "Maximum aliased size" setting - how it could be incorporated.
Side note: I'm not sure if I ever noticed or it's something relatively new:
In KDE Settings: Fonts: There is an "Exclude range from anti-aliasing" setting.
Ha! They stole a page from my book! That's what my setting does.
Although it's tailored for our particular use for one font. So it's not going away ;-)

PR #751, opened 2020-01-15
Remove font 0 from UI

kybos:
This removes font 0 from the settings and replaces it by the actual application font set by Qt style or in a stylesheet, as discussed in #750.


terminator356:
Thanks. We'll go from here and figure something out.

PR #752, opened 2020-01-17
New options for font settings

kybos:
As discussed in other places...
I think I found a way to enable scaling the fonts automatically without any incompatible changes. On the other hand, full customizability is still ensured, as the user can switch the new option off and set the font sizes individually as desired.

image

### DIFF
--- a/muse3/muse/app.cpp
+++ b/muse3/muse/app.cpp
@@ -1043,6 +1043,14 @@ MusE::MusE() : QMainWindow()
       MusEGlobal::config.fonts[0].setPointSize(font().pointSize());
       MusEGlobal::config.fonts[0].setBold(font().bold());
       MusEGlobal::config.fonts[0].setItalic(font().italic());
+
+      int fs = font().pointSize();
+      MusEGlobal::config.fonts[1].setPointSize(qRound(fs * MusEGlobal::FntFac::F1));
+      MusEGlobal::config.fonts[2].setPointSize(qRound(fs * MusEGlobal::FntFac::F2));
+      MusEGlobal::config.fonts[3].setPointSize(qRound(fs * MusEGlobal::FntFac::F3));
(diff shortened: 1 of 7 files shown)


terminator356:
Sorry for the delay.
Let's go with this for now. Probably the most reasonable solution for now.

We might be able to eliminate a few of those fonts.
For example the ones marked as 10, I think they are only used by one or two places,
 and they may be places that would want to naturally scale with app font.
I've never adjusted them. It seems reasonable that we might be able to
 eliminate them since they always say 10 anyways and likely could be
 simply made the same size as the app font.
We'll have to double check what uses them to see if that's possible...

kybos:
Yes, a good idea.
I will investigate.

kybos:
Font 2 is used for the Transport, but it only scales the labels:
image
That does not make much sense, does it?
Font 2 is the only one that was supposed to be equal to the system font in the default setting. Would it be a candidate for removal?

terminator356:
Ha! I suppose the idea was that users could make the transport smaller than the main font.
But if it is not being applied to the transport as a whole, it is useless.
Should we keep font 2 and apply it to the whole transport?
Hm... Hard to say. I mean, one could argue that we give mixer strips their own font,
 so why not remove the strip font as well... but there is a difference here:
The mixer strips are 'embedded' and there can be a lot of them in the mixer so I
 absolutely need that smaller font to see more strips. The transport on the other hand
 is a separate window and there's only one transport.
Given that it has been this way for a long time, I would say go ahead and remove font 2.
I don't think anyone will miss it. Maybe in the future if someone asks we could add it again,
 but for now, go ahead and remove it.
Any objections? Robert any opinion on this?

spamatica:
I trust your judgment on this.

Den ons 22 jan. 2020 kl 01:27 skrev Tim <notifications@github.com>:

PR #754, opened 2020-01-22
Fluidsynth - set soundfont for all/empty channels

kybos:
A function I was badly missing when using GM soundfonts...

With Ctrl the soundfont is set for all channels.
With Shift only for channels not yet filled (+ the currently selected one, of course).


terminator356:
Thanks.

PR #756, opened 2020-01-24
Fix event copy selection in piano roll (fixes #747)

### LINKED ISSUE #747: Ctrl+Drag in piano roll editor should only select new event
Small usability suggestion in MIDI editor/piano roll.

When I use the Ctrl key to make a copy of an event and drag it to another location, the original stays selected. I think only the new event should be selected.

My reason for this is that I usually start with a "live" MIDI recording, then drag around copies of the events and move and change the length of them. Now I have to de/reselect all the time.

This will also make it more consistent with other software, for example LibreOffice Draw.

muse3_ctrl_drag

kybos:
The selection only seems to work when set directly in this function.
All the later selection modifications for the dragged events in the higher level code don't have any effect.

### DIFF
--- a/muse3/muse/midiedit/prcanvas.cpp
+++ b/muse3/muse/midiedit/prcanvas.cpp
@@ -1012,7 +1012,9 @@ bool PianoCanvas::moveItem(MusECore::Undo& operations, CItem* item, const QPoint
       NEvent* nevent = (NEvent*) item;
       MusECore::Event event    = nevent->event();
       int npitch     = y2pitch(pos.y());
+      event.setSelected(false);
       MusECore::Event newEvent = (dtype == MOVE_COPY) ? event.duplicate() : event.clone();
+      newEvent.setSelected(true);
       int x          = pos.x();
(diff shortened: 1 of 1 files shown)


terminator356:
I guess this in reference to the user request a few weeks ago? Thanks.
I can see how the old behaviour could annoy when re-dragging an event.

Tip: Florian expressed a desire to have all selections undo-able.
He added that to his score editor, and even parts and events if I recall.
But I felt that undo should only be for things which have been edited.
I felt that undo 'backpedaling' through possibly many selections, to get to the last edit,
 seemed much overkill and could likely annoy.
So I reversed what he did but I left the door open for possible user choice in the matter.
IIRC I added a new config item 'selections undoable' but I have not exposed it yet to the user.
It remains 'off' for now.
In the code, I have mostly tried to honour the setting, in his code and elsewhere.
The effort may not be complete but I think in the places that I reversed his work, I tried
 to replace it with that setting (I didn't want to completely erase what he had done).

PR #761, opened 2020-01-29
Fix custom columns handling

kybos:
While I was at the header fixes I noticed that there were some open issues and work in progress related to the custom columns. I tried to finish what was apparently originally intended. I hope I did not miss anything critical.

1. Clean up the two duplicate vectors (custom_columns vs new_custom_columns) and related code.
2. Enable dynamic adding/removing of custom columns (without the need to restart the complete app, as it was the case before).
3. Allow the user to cancel the Custom columns modal dialog, without applying the settings (not possible before).

### DIFF
--- a/muse3/muse/arranger/arranger.cpp
+++ b/muse3/muse/arranger/arranger.cpp
@@ -79,8 +79,7 @@
 
 namespace MusEGui {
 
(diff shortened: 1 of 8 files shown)


terminator356:
That was Florian's work. I've never used it and have no idea if it's still working.
Hopefully this allows it to live on, for those who use it. Thanks!

PR #766, opened 2020-02-11
Improve mixer strip display button

kybos:
It's that button again... It always looked so bad, not matching the (varying) width of the mixer strip and looking to misplaced down there...
So this is another idea: Make it flat, small and asymmetric by design, so it does not look like a mistake.
I added a keyboard shortcut and different mouse cursor, so it's clear there is an action behind it.
IMO it looks quite nice and clean this way...

image

BTW: The mixer strip is also displayed in the midi/wave editors, but there it can't be hidden.

### DIFF
--- a/muse3/muse/arranger/arranger.cpp
+++ b/muse3/muse/arranger/arranger.cpp
@@ -367,11 +367,14 @@ Arranger::Arranger(ArrangerView* parent, const char* name)
       trackInfoButton  = new CompactToolButton(this);
       trackInfoButton->setContentsMargins(0, 0, 0, 0);
       trackInfoButton->setIcon(*mixerstripSVGIcon);
-      trackInfoButton->setToolTip(tr("Show mixer strip for current track"));
+      trackInfoButton->setToolTip(tr("Show mixer strip for current track (F5)"));
       trackInfoButton->setCheckable(true);
       trackInfoButton->setChecked(showTrackinfoFlag);
       trackInfoButton->setFocusPolicy(Qt::NoFocus);
-      trackInfoButton->setFixedSize(52, 14);
+      trackInfoButton->setFixedSize(20, 14);
+      trackInfoButton->setDrawFlat(true);
+      trackInfoButton->setShortcut(QKeySequence(Qt::Key_F5));
+      trackInfoButton->setCursor(QCursor(Qt::PointingHandCursor));
       connect(trackInfoButton, SIGNAL(toggled(bool)), SLOT(showTrackInfo(bool)));
 
       genTrackInfo(trackInfoWidget);
@@ -1097,7 +1100,7 @@ void Arranger::controllerChanged(MusECore::Track *t, int ctrlId)
 
 void Arranger::showTrackInfo(bool flag)
       {
-      trackInfoButton->setToolTip(flag ? tr("Hide mixer strip for current track") : tr("Show mixer strip for current track"));
+      trackInfoButton->setToolTip(flag ? tr("Hide mixer strip for current track (F5)") : tr("Show mixer strip for current track (F5)"));
       showTrackinfoFlag = flag;
       trackInfoWidget->setVisible(flag);
       updateTrackInfo(-1);
--- a/muse3/svg/mixerstrip.svg
+++ b/muse3/svg/mixerstrip.svg
@@ -1,44 +1,66 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<svg id="SVGRoot" width="64px" height="64px" version="1.1" viewBox="0 0 64 64" xmlns="http://www.w3.org/2000/svg" xmlns:cc="http://creativecommons.org/ns#" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
- <defs>
-  <linearGradient id="linearGradient1637" x1="31" x2="31" y1="4" y2="61" gradientTransform="matrix(.25 0 0 .80702 -2.75 5.7719)" gradientUnits="userSpaceOnUse">
-   <stop stop-color="#f00" offset="0"/>
-   <stop stop-color="#ffd42a" offset=".29598"/>
-   <stop stop-color="#008000" offset=".5068"/>
-   <stop stop-color="#008000" offset="1"/>
-  </linearGradient>
-  <linearGradient id="linearGradient1655" x1="31" x2="31" y1="4" y2="61" gradientTransform="matrix(.25 0 0 .80702 15.25 5.7719)" gradientUnits="userSpaceOnUse">
-   <stop stop-color="#f00" offset="0"/>
-   <stop stop-color="#ffd42a" offset=".29598"/>
-   <stop stop-color="#008000" offset=".5068"/>
-   <stop stop-color="#008000" offset="1"/>
-  </linearGradient>
-  <linearGradient id="linearGradient1657" x1="31" x2="31" y1="4" y2="61" gradientTransform="matrix(.25 0 0 .80702 33.25 5.7719)" gradientUnits="userSpaceOnUse">
-   <stop stop-color="#f00" offset="0"/>
-   <stop stop-color="#ffd42a" offset=".29598"/>
-   <stop stop-color="#008000" offset=".5068"/>
-   <stop stop-color="#008000" offset="1"/>
-  </linearGradient>
-  <linearGradient id="linearGradient1659" x1="31" x2="31" y1="4" y2="61" gradientTransform="matrix(.25 0 0 .80702 51.25 5.7719)" gradientUnits="userSpaceOnUse">
-   <stop stop-color="#f00" offset="0"/>
-   <stop stop-color="#ffd42a" offset=".29598"/>
-   <stop stop-color="#008000" offset=".5068"/>
-   <stop stop-color="#008000" offset="1"/>
-  </linearGradient>
- </defs>
- <metadata>
-  <rdf:RDF>
-   <cc:Work rdf:about="">
-    <dc:format>image/svg+xml</dc:format>
-    <dc:type rdf:resource="http://purl.org/dc/dcmitype/StillImage"/>
-    <dc:title/>
-   </cc:Work>
-  </rdf:RDF>
- </metadata>
- <g fill-rule="evenodd" stroke="#000" stroke-dashoffset="37.795" stroke-linejoin="round">
-  <rect x="1" y="9" width="8" height="46" fill="url(#linearGradient1637)" style="paint-order:fill markers stroke"/>
-  <rect x="55" y="9" width="8" height="46" fill="url(#linearGradient1659)" style="paint-order:fill markers stroke"/>
-  <rect x="37" y="9" width="8" height="46" fill="url(#linearGradient1657)" style="paint-order:fill markers stroke"/>
-  <rect x="19" y="9" width="8" height="46" fill="url(#linearGradient1655)" style="paint-order:fill markers stroke"/>
- </g>
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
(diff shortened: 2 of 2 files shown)


terminator356:
Ah that pesky button, eh?
The other editors can hide the strip with the splitter although there's no button for that,
 there's not a whole lot of point to spending time adding one...
Thanks.

kybos:
You are of course right about the splitter.
So perhaps it would be better to remove this button completely? For the sake of consistency between the views, at it would certainly look better without, no matter what I do with it...

terminator356:
Well, there's a slight problem. If the splitter is like, way over there,
 the user must drag it all the way back to hide the strip.
Worse, it does not remember where it was.
The button allows users to flip the trackinfo panel on and off while retaining its size.

If anything, we might add this to all the editors.
However there are some tricky issues with where to put it I believe.
The editors don't have that extra space at the bottom extending all the way to the left.
It would require fairly serious layout rearrangements in each editor.

And then there's that pesky Ctrl button. I've always hated where it is.
In the Pianoroll it's OK but in the Drum Editor it really looks stupid way over on the left
 when the drum list is expanded to a reasonable normal size.
What else can we do with that thing... Put it on a toolbar?
Look closely, we need to be real creative here.

kybos:
OK, we can leave it as it is, it was just a question... It looks acceptable now, especially on the darker themes.
I also thought about putting the button in the toolbar, but did not find a good place there either (without creating a new separate toolbar for it alone).

I personally have no use for hiding the mixer strip or changing its size. Perhaps hiding makes sense for small laptop screens. But changing the width breaks the layout, many items are not properly resized and it looks more like a mistake, rather designed to be fixed width. 
I am probably missing something essential, but what's the use of changing the width at all? For potentially elided texts etc. there are tooltips, what else?

terminator356:
One huge reason: Text.
Want to see more text in a track name? Or a midi strip control name?
It's the only way to display horizontal text in a normally narrow control.
And in several of them at once, at a glance instead of hovering for tooltips.
To me it made perfect sense. Turned out to be the perfect solution to a problem
 that plagued us. Our strips were fixed size before. Text could not be distinguished
 from other text. That's very important. Wait till you see the other stuff I'm planning in there.
Originally I had implemented putting those purple extra midi strip controls into 
 an extra side panel to the right of the strip. But I decided to put them along with 
 the blue controls into two 'tabs'. Still have visions and support in code for some of the
 ideas. (Hedging my bets in case I need it.)

Ardour also has a strip expander.

What's this about breaking layout when expanded? Pictures?
Do you mean the textual controls in a strip expand but the others do not?
Intentional. No need to expand the others. Only the ones with potentially long text.

kybos:
OK, I see. I think that was what I meant.
Do you know what this pane is in the expanded area (marked with red ellipse on the screenshot)?
It's only visible in the audio strip, not in midi.

image

terminator356:
Oh wow. That shouldn't show up. Supposed to be hidden for now. I'll take a look.
Here you see a bit of what I was planning and talking about.
I was going to put stuff in there. If I ever get around to it.

terminator356:
Hi. It's hidden for now. Can you test and see if anything looks unusual,
 apart from the unusual design ;-)
I recall some weirdness years ago when it was hidden, but it tests OK here.

I hope you don't mind but I replaced your magnet cursor with the pencil in the controller view.
It just seemed like we needed better pinpoint accuracy there.

May I ask, how do the actual drum notes (diamonds) look on your display?
Too small? Could you use another zoom level?

Favour to ask: Can you do a quick test to see how my new 'note move' cursor looks?
It replaces the standard awful 4-pole arrow which was badly blocking the view
 of those small diamonds.
Grab a drum note and move it around to see the cursor.
I know, the offset needs a little adjustment. First 'draft', he he.
The 4 arrows of the new move cursor should 'hug' the diamond comfortably
 but not be say, way out there.
In the icons file when creating it, I had to multiply the size value by 2.
Does it work OK on your screen? Thanks.

spamatica:
Den lör 15 feb. 2020 kl 04:10 skrev Tim <notifications@github.com>:

> Hi. It's hidden for now. Can you test and see if anything looks unusual,
> apart from the unusual design ;-)
> I recall some weirdness years ago when it was hidden, but it tests OK here.
>
> I hope you don't mind but I replaced your magnet cursor with the pencil in
> the controller view.
> It just seemed like we needed better pinpoint accuracy there.
>
Yes. the magnet was not working very well, I prefer it this way.

A related bug that has been there forever is that the icon does not change
if you change to pencil after the mouse has entered the controller
automation area. Being a heavy shortcut user one hand always rests on the
keyboard shortcuts.
You can try it by pressing A up in the editor, move the mouse to the
controller area and press D, the icon does not change, toolbar changes
though.

May I ask, how do the actual drum notes (diamonds) look on your display?
> Too small? Could you use another zoom level?
>
Drum notes are fine here.

> Favour to ask: Can you do a quick test to see how my new 'note move'
> cursor looks?
> It replaces the standard awful 4-pole arrow which was badly blocking the
> view
> of those small diamonds.
> Grab a drum note and move it around to see the cursor.
> I know, the offset needs a little adjustment. First 'draft', he he.
> The 4 arrows of the new move cursor should 'hug' the diamond comfortably
> but not be say, way out there.
> In the icons file when creating it, I had to multiply the size value by 2.
> Does it work OK on your screen? Thanks.
>
Yeah, that's not a bad idea. The actual icon might need some work though.
It might be an optical illusion but it looks out of focus compared to the
drum notes, is it scaled somehow?

kybos:
> Hi. It's hidden for now. Can you test and see if anything looks unusual,
> apart from the unusual design ;-)
> I recall some weirdness years ago when it was hidden, but it tests OK here.

Looks OK here too.

> May I ask, how do the actual drum notes (diamonds) look on your display?
> Too small? Could you use another zoom level?
 
They are relatively small, but it's OK, no issue here.
There are 2 zoom levels now, but the diamonds remain the same size in both.
What other zoom level did you have in mind? Double the view size, including the diamonds and the texts?

> Favour to ask: Can you do a quick test to see how my new 'note move' cursor looks?
> It replaces the standard awful 4-pole arrow which was badly blocking the view
> of those small diamonds.
> Grab a drum note and move it around to see the cursor.
> I know, the offset needs a little adjustment. First 'draft', he he.
> The 4 arrows of the new move cursor should 'hug' the diamond comfortably
> but not be say, way out there.
> In the icons file when creating it, I had to multiply the size value by 2.
> Does it work OK on your screen? Thanks.

The new cursor is too big here (it partly covers the sound names in the tooltips).
It does not move ("jump") with the diamond, sometimes there is an offset (as you mention above).
There is a third (double-arrow) cursor displayed when mouse is over the note, should it be also replaced?
The colour could be perhaps too dark for a dark background.
The new cursor does not appear when you create a new note a move it immediately.

kybos:
> And then there's that pesky Ctrl button. I've always hated where it is.
> In the Pianoroll it's OK but in the Drum Editor it really looks stupid way over on the left
> when the drum list is expanded to a reasonable normal size.
> What else can we do with that thing... Put it on a toolbar?

You are right, it looks quite confusing, at the first sight I thought the arrow belonged to the button.

image

It could really be better to put it on the toolbar or in the menu and add a keyboard shortcut for fast access (same for the mixer strip hide button).

kybos:
What about adding a plus (and capitalize, as it should be anyway):

image

Looks a little better to me, more expressive, not so lost down there...

terminator356:
All three of the Ctrl, and 'S' and 'X' buttons on the controller panel, need to be icons.
With tooltips of course, describing what they do.
Long overdue. Been planning to do it. What I had in mind:

All three button icons start with the same base icon:
Preferably a miniaturised version of our round controls.
Or if you can think of some other icon meaning 'control' as in knob or slider etc.

For the Ctrl icon we include a small red 'plus' sign in the top right.
Or, one of those 'sparkly spark' things often found at the top right of 'new document' icons,
 meaning 'fresh' or 'new'.

For the 'X' button, we include a red circle-with-line-through much like your new panic icon.

For the 'S' button, we do nothing more than use the base icon, since it's meaning
 is "select control".

How's that work?

BTW It is very tempting to redo the drum editor layout so that the left edge of the
 scrollbar lines up with the left edge of the actual canvas.
Very tempting indeed.
But unfortunately, the drum list is so wide that many times the canvas will be
 showing as rather skinny in relation to the list.
So we want the advantage of the full scrollbar width so that we are not left
 manipulating a very short scrollbar.
I didn't design it that way, but you can see why it was done.

terminator356:
> tempting to redo the drum editor layout so that the left edge of the
scrollbar lines up with the left edge of the actual canvas

Ha! Ironically that's what we do with the Arranger.
Hm, I wonder if we should ever change that to mimic the drum editor's way?
I must have considered it before... or not.
[ Sigh... ] MusE is full of contradictions, eh? Midi vs. audio, waves vs. notes...

kybos:
> All three of the Ctrl, and 'S' and 'X' buttons on the controller panel, need to be icons.
> With tooltips of course, describing what they do.
> Long overdue. Been planning to do it. What I had in mind:
> 
> All three button icons start with the same base icon:
> Preferably a miniaturised version of our round controls.
> Or if you can think of some other icon meaning 'control' as in knob or slider etc.
> 
> For the Ctrl icon we include a small red 'plus' sign in the top right.
> Or, one of those 'sparkly spark' things often found at the top right of 'new document' icons,
> meaning 'fresh' or 'new'.
> 
> For the 'X' button, we include a red circle-with-line-through much like your new panic icon.
> 
> For the 'S' button, we do nothing more than use the base icon, since its meaning
> is "select control".
> 
> How's that work?

That sounds good, I fully agree. With the "round controls" you mean the knob, as in the mixer strip?

> BTW It is very tempting to redo the drum editor layout so that the left edge of the
> scrollbar lines up with the left edge of the actual canvas.
> Very tempting indeed.
> But unfortunately, the drum list is so wide that many times the canvas will be
> showing as rather skinny in relation to the list.
> So we want the advantage of the full scrollbar width so that we are not left
> manipulating a very short scrollbar.
> I didn't design it that way, but you can see why it was done.

Maybe that was the reason long time ago, when screens were small and low resolution ;-).

Still it's a design error (or at least very non-standard, to put it euphemistically ;-) ), because a scrollbar is supposed to be restricted to the UI element it moves, which in this case is the drum canvas and nothing else. I don't think you will find anything like that in another application. It's confusing (especially in connection with the ctrl button in our case), and it's also inconsistent with the other views in MusE.

terminator356:
Redesigned the drum note move cursor. It's now a half pencil, half 4-way arrow.
Seems to work well here so far for a first try. I also fixed the mouse-over-item cursor.
This time in icons.cpp when I create it, it has the exact same parameters as the pencil cursor.
So it should look like all the other cursors especially the pencil.
@kybos How does it look over there?
Here's what it's supposed to look like:
drum new move cursor 1

kybos:
> Redesigned the drum note move cursor. It's now a half pencil, half 4-way arrow.
> Seems to work well here so far for a first try. I also fixed the mouse-over-item cursor.
> This time in icons.cpp when I create it, it has the exact same parameters as the pencil cursor.
> So it should look like all the other cursors especially the pencil.
> @kybos How does it look over there?

The new cursor looks nice, a good solution.

Some points I mentioned above are still valid:
- If you create a new note and move it immediately, the cursor does not change (or is it intentional?).
- The arrow part of the icon will be difficult to see on dark backgrounds (probably not so bad).
- When you hover over a note, the cursor changes to the new one, when you press the mouse button, it changes back to the normal pencil (before you start moving) - intention or not?

terminator356:
Slight redesign. Inner part is light grey. Difficult to see anything at that size,
 but hopefully looks better on a dark background.
Still not happy about arrows, difficult to see again, should enlarge slightly.

The other issues... meh... part intentional, part neglect, part consequence of other stuff
 over the years... Hard to say. I'll see if there's time to address them.

PR #769, opened 2020-02-16
harmonize VST support in UI

spamatica:
Here's a stab at fixing some VST inconsistencies.
It might have sounded dramatic in the mail and made to look bigger than it is. I have made only changes to what is visible in the UI (and to me that is fine). 

- Plugin path dialog only has one type of VST plugins now: VST. Config still stores both kinds.
- Right click menu has been relabeled VST instead of Native VST
- Wine VST from the plugin filter dropdown has been removed.

### DIFF
--- a/muse3/ChangeLog
+++ b/muse3/ChangeLog
@@ -1,3 +1,7 @@
+16.02.2020:
+    - Removed the VST tab referencing Wine VST and relabeled Native VST as VST (rj)
+    - Remove Wine VST plugin type in dropdown as we only support that through
+      dssi-vst and it is dead since many years
 15.02.2020:
     - Fixed issue with plugin scan #763, where a valid vst plugin but nonfunctional was never
       added to the cache and hence caused a rescan every time (rj)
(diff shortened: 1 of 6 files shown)


kybos:
From my perspective an important usability improvement, I was confused by the different VSTs as well.

terminator356:
I agree with the changes except for one thing, if I understand correctly:
The name should be more specific: Linux VST.
Simply calling it VST gives the impression that we support VSTs of all kinds.
I think that may cause confusion to a user.
You could argue that should the time come that we get built-in support for Win VSTs,
 say when we add built-in VST3 support, we'll create the necessary tab for it
 and call it say 'Win VST' or 'VST3'. True, fair enough.
But then this new name 'VST' you've chosen would have to be renamed 'Linux VST' anyways.
Follow? Can you convince me to leave it at 'VST'? 'Cause I think it ought to be changed.
I suppose you could argue that calling it 'VST' by de-facto will mean 'Linux VST',
 and any other tabs we add in the future will be more specific. Fair enough.
But still, 'VST' a bit too vague I think... Thanks.

kybos:
I checked in Ardour, QTractor and Carla, and I see only VST in most places.
In Carla (which I think is the only one with VST3 support), there is a distinction between VST2 and VST3 in the detailed plugin list.
But possibly the concepts are different...

terminator356:
Stop the presses! Exciting developments seem to have fixed dss-vst.
Continuing that thread in the team area, check it out:

PR #780, opened 2020-03-16
Piano roll multi resize (fixes #748)

### LINKED ISSUE #748: Feature request: Change length of all selected notes
Piano roll. When I have several selected notes and edit the length with the Pencil (by dragging), the length of all selected notes should be affected/adjusted. Now only the note that I change, gets changed.

kybos:
Hi guys,
Please check this PR. I tested everything as far as possible, also the many corner cases, so I hope I did not break anything (and even corrected some bugs in the item resizing). 
Function: In piano roll, edit mode (pencil), when Ctrl is held all selected items are resized.
Thanks.

### DIFF
--- a/muse3/ChangeLog
+++ b/muse3/ChangeLog
@@ -1,3 +1,18 @@
+18.03.2020
+	- Enable multiple resizing in piano roll (#748) (kybos)
+	- Enable CTRL+Left Mouse Click item selection in edit mode in piano roll (kybos)
+17.03.2020
+	- Add more information to note tooltips in midi editors (kybos)
+	- Remove fixed track heigth from file templates, to enforce global setting (kybos)
+09.03.2020
+	- Transfer Help content to github wiki, link Manual to wiki in MusE (kybos)
+08.03.2020
+	- Fix EditPaste cursor HiDPI issue in midi editors->controller views (kybos)
+05.03.2020
+	- Fix tool cursor not changing in midi editors->controller views (kybos)
+	- Fix background color setting ignored in midi editors->controller views (kybos)
+04.03.2020
+	- Add tool context menu to Score Editor (kybos)
 28.02.2020:
     - LV2: Now shows MidiName patch names (gmsynth.lv2 for example). (Tim)
     - Added share/rdf/README to describe what's there and why.
--- a/muse3/muse/components/canvas.cpp
+++ b/muse3/muse/components/canvas.cpp
@@ -914,7 +914,11 @@ void Canvas::viewMousePressEvent(QMouseEvent* event)
                                   setCursor();
 
                                   if (supportsMultipleResize) {
-                                      selectItem(curItem, true);
+                                      if (!curItem->isSelected()) {
+                                        deselectAll();
+                                        deselect_all = true;
+                                        selectItem(curItem, true);
+                                      }
                                       if (resizeDirection == RESIZE_TO_THE_RIGHT)
                                           resizeSelected(start.x() - curItem->x() - curItem->width());
                                       else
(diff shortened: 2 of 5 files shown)


spamatica:
cool, will try it tonight

spamatica:
Did a first test. First I was stumped and saw no change at all. Looked in the code and I see it requires CTRL. Works! I suppose you wanted to preserve the old functionality though is that necessary?
When you have multiple notes selected it feels natural to me that they all are affected.

spamatica:
Ah, seems there are some side effects in the arranger. The resizing affects parts too but not in a good way. Looks OK while you are dragging but when you let go it gets a little crazy.
I wonder if there is a use case in the arranger or if we can prohibit this functionality there?

kybos:
Thanks, will check tomorrow. It’s not supposed to affect the arranger.

kybos:
> First I was stumped and saw no change at all. Looked in the code and I see it requires CTRL.

I wrote it above ;-) :
"Function: In piano roll, edit mode (pencil), **when Ctrl is held** all selected items are resized."

> I suppose you wanted to preserve the old functionality though is that necessary?

Yes, in fact I did it without CTRL first, but then thought the old behaviour should be kept for backward compatibility and consistency with the part canvas.
I can change it back if you think it's not necessary. That would allow for using Ctrl+click for selecting the items, which would be very practical in my eyes, and even consistent with the drum editor.

> seems there are some side effects in the arranger.

This should be corrected now. I did not notice it because it does not happen when in debugger, interestingly...

spamatica:
Den tis 17 mars 2020 kl 10:41 skrev kybos <notifications@github.com>:

> First I was stumped and saw no change at all. Looked in the code and I see
> it requires CTRL.
>
> I wrote it above ;-) :
>
Hehe, okay, goes to show I can't read :D

"Function: In piano roll, edit mode (pencil), *when Ctrl is held* all
> selected items are resized."
>
> I suppose you wanted to preserve the old functionality though is that
> necessary?
>
> Yes, in fact I did it without CTRL first, but then thought the old
> behaviour should be kept for backward compatibility and consistency with
> the part canvas.
> I can change it back if you think it's not necessary. That would allow for
> using Ctrl+click for selecting the items, which would be very practical in
> my eyes, and even consistent with the drum editor.
>

Sounds good to me.


> seems there are some side effects in the arranger.
>
> This should be corrected now. I did not notice it because it does not
> happen when in debugger, interestingly...
>
Wow, wierd. Some timing issue maybe?

Reminds me of the mixer, I did a bug report a while ago.
It happens to me quite often that strips do not appear there. My primary
suspect is a timing issue when MusE is under heavy load...

kybos:
OK, it's changed now as discussed.
All selected items (+current one if not selected before) are resized by default.
Ctrl+Left click toggles item selection (+Shift toggles all items of same pitch).

Can you please check again?

Later I will add the possibility to select exactly one item for editing by right mouse click, since this is not so easily possible any more (without using a keyboard shortcut to Deselect all or switching to pointer mode).

spamatica:
Seems to work now!

The only thing that is a little bit weird is if you have selected a note and then start dragging on another note that is not selected, the new note will be selected and both will be edited.

Maybe the solution is to deselect the other note(s) before starting to drag the new one, in my head that feels logical. Probably not a big issue. Otherwise it seems to work fine.

kybos:
Changed. I was wondering myself which alternative was more logical...
I will merge the PR now, OK?
Tim seems not available ATM, I hope he is well.

spamatica:
Yes, fine by me.

PR #788, opened 2020-03-21
added extend end of wave part

spamatica:
Added the ability to extend wave parts if there is more audio after.

Next up is extending to the left, which is a bit more tricky...

### DIFF
--- a/muse3/ChangeLog
+++ b/muse3/ChangeLog
@@ -1,5 +1,6 @@
 21.03.2020
     - Add shortcut (Ctrl+U) for opening synth UI for selected synth-track in arranger (rj)
+    - Added wave part extend beyond end of part if there is more wave file (rj)
 18.03.2020
 	- Enable multiple resizing in piano roll (#748) (kybos)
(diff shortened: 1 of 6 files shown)


terminator356:
Hi. What does this do exactly?
We already have the ability to extend parts to the right, AND left (Flo added that I think).
Don't forget we have those special 'arrow' graphics drawn at the right end of a part
 to indicate there is 'more content' that can be shown.

spamatica:
As far as I know it has never worked correctly, it was possible to extend
the part, but the actual audio event was not extended so it wasn't really
doing anything useful.

If I manage I'd also I'd like to change the way extension to the left
works. Applies to midi too. It isn't really extending, it is moving the
part left and making it longer. Instead the notes should stay in the spot
they were meant to be and the beginning should be extended.

Den lör 21 mars 2020 kl 23:28 skrev Tim <notifications@github.com>:

terminator356:
Right. OK Thanks. Carry on.
Good luck, keep your thinking cap on, I know there will be tricks and traps...

PR #798, opened 2020-04-19
Add Python 3 compatibility to scripts.

dfandrich:
They should now be compatible with at least Python 2.7 and Python 3.8.


terminator356:
Great stuff, thanks!
When I redesigned the Python stuff several months ago (ChangeLog 27.08.2019) I noted
 that it was mostly all converted over to be compatible with 2/3, except that "the bundled
 midi scripts ('Functions') require Python 3" (see README). I could not think of a way,
 or didn't really have the time (?), to make those scripts 2/3 compatible.

And it seems I missed these util scripts and a mistake in the launcher, eh?
Great that it should be 2/3 compatible.
Thanks again.

PR #800, opened 2020-04-26
Rj development

spamatica:
Further work after help from Tim. Seems to be working passably now.

Basically, extending parts to the left and right will reveal more data if more is available and leave the part events in the same absolute position in the song. Audio parts cannot be extended beyond the beginning of the audio event as it does not make sense.

Redrawing audio on the fly when extending to the left is not implemented as it made my brain melt.


kybos:
Good stuff, this change really makes sense.
What I noticed: There is this triangle displayed on the right end of the part when events are cut off on the right. Now with the new logic events can also be cut off on the left - shouldn't the triangle be displayed on the left end in this case? It is now displayed on the right side, and also not quite consistently - I noticed that it's not displayed at all if only the first couple of items in the part are cut off.

terminator356:
I too wondered this when looking at Robert's work, what about the left side 'triangle'...
When I added that 'triangle' thing to indicate that there are 'hidden' events
 not shown to the right, I kind of, sort of, left the door open for the left side as well,
 should the day come when we needed it (now).
I think it mostly is 'ready' but may need some further work of course.
I can help there...

terminator356:
Something else just occurred to me:
How is it possible to have events that are before a part's left border,
 when both our part and event class use unsigned ints for the tick value?
That would make it impossible to have events with negative tick values, no?

spamatica:
Yeah, that triangle, I will see if I can add it for the left side too. 

As for ticks before the left border, a bit of a corner case, you are probably right they can't be negative. It can happen though that the left has been adjusted to the right, covering up events.

spamatica:
Wave files before the start of a part is another story, that happens a lot when you cut out parts in the middle of a take. The event is usually at zero and spos (wave file offset) points to the actual position of the wave in the part so nothing is negative here either.

terminator356:
> It can happen though that the left has been adjusted to the right, covering up events.

That's the part I'm scratching my head over - how can a part's left side be 'covering'
 events when the events can't have negative ticks indicating they are before the left side?

terminator356:
Hi, Sorry for the delay here.
I have not yet tried the code or examined it, but after reviewing our existing code I
 concluded it is not possible at the moment to have events that are before the left border of a part.
Therefore the whole 'left arrow' idea is also a no-go at the moment.

This is because all timing values, both frames and ticks, currently use unsigned integers.
It may be tempting to 'subtract' values from the times so that the resulting value
 is like 0xffffffff or less, but the rest of the program will not like that, it will think those are
 legitimate time values waaay at the end of the part.

Theoretically, 32-bit unsigned integer frames gives us a maximum recording time of
 27 hours at 44100 Hz. Even Jack's own frame times are unsigned integers.
But, it is unwise to trust that the whole application and the various libraries and
 system functions that it talks to, will be able to handle the range of unsigned integers
 from 0x80000000 to 0xffffffff. Even some of our code converts to signed integers
 for very easy relative offset calculations. (Actually that is not the correct way, it IS
 possible to do such calculations while preserving the unsigned range of the target,
 but it is more complex requiring a conditional check of the offset's polarity, but it
 is a technique which I always try to use in my code, just for range completeness.)
For example: Qt pixel graphic functions use signed integers.
From that alone, it seems we can't really rely on unsigned integers.

If we make our time values unsigned, that is the most guaranteed compatible with
 all other parts of the system.
But it halves the maximum recording time (frame). We discard fully **one half** of
 Jack's maximum frame time simply because we can't display or edit the data, for example.
But again, it's questionable whether the app could **fully** correctly deal with the
 upper half of the range anyway.

Solutions:
1) For Robert's new feature, when moving a left border to the right, if any notes
would require a negative time value (meaning they _would_ be before the part's left border),
 simply 'bump' them over to the right - that is, force their time values to be zero.
Kind of destructive since the original values are lost and expanding to the left again
 will not restore them. 

2) Keep our unsigned integers, but in all critical areas that work with the values,
 treat them as signed - consistently. That is, values between 0x80000000 and 0xffffffff
 are to be treated as negative. This is a lot of work, and would just likely
 dig us into a deeper hole, because the real solution is 3:

3) Everything everywhere must be changed to use signed integers.
**Ultra-massive** job. But the 'right thing to do'.
To that end, several merges ago I created "muse_time.h" which simply has:
`typedef int64_t MuseFrame_t;`
`typedef int64_t MuseCount_t;`
I began to use it in new code but I admit, keeping to the plan has been tough,
 and I neglect it sometimes.

And finally, there is a comment in arranger.cpp, although not directly related,
 it illustrates the kind of problem we face, that everything has limits,
 and some limits impose even lower limits on seemingly wide-range values:
`// song length is limited to 10000 bars; the real song len is limited`
`// by overflows in tick computations`
`lenEntry = new SpinBox(1, 10000, 1);`

Cheers, all for now...

spamatica:
Den sön 24 maj 2020 kl 22:41 skrev Tim <notifications@github.com>:

> Hi, Sorry for the delay here.
> I have not yet tried the code or examined it, but after reviewing our
> existing code I
> concluded it is not possible at the moment to have events that are before
> the left border of a part.
> Therefore the whole 'left arrow' idea is also a no-go at the moment.
>
> Yeah, I had figured as much.

Keeping the events that are before _does_ work, though, probably only by
coincidence.
In the running session the events are still there so they will have valid
position values when the part is adjusted so they are "inside" the part
again. That could also be the reason why the triangle is displayed on the
right side, as the events will be located some 4 million frames into the
future.
When the song is stored and reloaded these events are gone..

So, waiting for the "right" fix in extending frames to signed 64bit maybe
I'll go ahead with the pull request anyway? It does improve usability in
adjusting part lengths, especially for audio parts. I'll give it some more
testing so there isn't any crashiness though.


This is because all timing values, both frames and ticks, currently use

terminator356:
I would say yes, if you're confident ;-)
Afterwards perhaps we can work on solution 1 (bumping notes to time zero)
 while the rest of the app waits for signed ints.

PR #812, opened 2020-05-23
Added new color scheme

kybos:
Hi guys,
This one is a little bigger, it has been my experimental branch where I tried to make the UI just a little more to my liking (with varying success...), at the same time keeping 100% compatibility with the existing look and style. So there are many new colours and css properties to make the GUI more customizable, also for people like me that don't like the retro look with the pseudo 3d elements, colour gradients etc. This also required several code changes, but I only changed things where I was quite sure that I can keep them compatible with the current behaviour.
Please check if it breaks something for you, optimally you shouldn't notice any difference.
Thanks.


spamatica:
Compiles, opens, no immediate problems at least.
Though I can't see much difference. :P
Browsing the changes it looks like there is some optional 3d-look. Or rather the new flat is probably the optional one. But I couldn't find where to control this?

kybos:
Sorry, I didn't mention that, my first priority was not to change/break the existing look.
There is a new theme (Dark Flat), which should be selectable from the Appearance dialogue (Styles->Colour schemes). If it's not there, then something is missing.

spamatica:
Ah yes, I'm stupid, I didn't actually checkout your code so there were no differences. :P 

Did a quick test now, not bad!
There is however a redraw issue on my system when switching between tracks.
I recorded a little video and attached it. Never mind the colors, OBS seems to have a mind of it's own regarding the colors...
[2020-05-25 14-00-41.zip](https://github.com/muse-sequencer/muse/files/4677079/2020-05-25.14-00-41.zip)

kybos:
Thanks for the video, will look into it (unfortunately it does not happen on my systems). 
It seems there are differences in size hint handling between Qt versions, I had similar problems in other contexts as well...

spamatica:
Tried it on another machine, same OS, Kubuntu 18.04, and it behaves the
same.
This one has  Qt 5.9.5, I can check the other one later.
Maybe I can lend a hand at trying to isolate the offending resize, later
though



Den mån 25 maj 2020 kl 14:25 skrev kybos <notifications@github.com>:

terminator356:
Eek. I'm out of disk space with all these 1GB MusE builds ;-)
Can't test yet. Trying to free up some space...
Should I assume the flickering only happens here with this PR, not with existing builds?

kybos:
Thanks, there is no hurry.
I was able to reproduce the problem myself (it occurs only sporadically and is quite fast on my machine).
It's no real flickering - when track is changed in the list, then the mixer strip is shortly resized to something slightly wider and then adjusted to the normal size. It certainly should not be the case with tracks of the same kind.
It happens only with this PR, it's not a general problem.

spamatica:
bild
Tried your latest fixes and the issue seems to be gone now.
Overall I like it very much. 
I have some trouble distinguishing which is the currently selected track though (attached a screenshot)
The color should be stronger or add some other visual cue to replace the gradient.

I noticed that the shortcuts no longer work in the mixer but I think that was broken already before this pr (I will double check)

Get this in and maybe we should drop a 3.2 release? :)

kybos:
Thanks for retesting.
I found workarounds for the erratic size hints (it's the slider scale causing it as I found out after a long search...), but I am still struggling with some detail issues, it's not perfect yet. And I have to resolve the conflicts with your recent PR, too.

The colours should look like that:
image

The selection should be bright yellow, and multiple selection light yellow.

If the colours are not loaded automatically (that seems to happen when switching between this branch and master, and keeping the config file), then you can load the correct ones manually in Appearance->Colours..., from the file /usr/share/.../themes/Dark Flat.cfg).

terminator356:
I am trying to build but get a problem:
`/usr/lib64/gcc/x86_64-suse-linux/9/../../../../x86_64-suse-linux/bin/ld:`
`../../muse/widgets/libmuse_widgets.a(meter.o): in function`
`MusEGui::Meter::setPrimaryColor(QColor const&)':`
`/home/me/Source/muse_20200528_kybos/muse3/muse/widgets/meter.cpp:531:`
`undefined reference to MusEGlobal::config`

@kybos Please remember that items in the widgets folder are to be completely
 independent, not having any references to global variables. (We went through
 this problem some time ago with the icons unit when a global variable was used.)
Try to give the meter widget its own primary colour variable with getter and setter methods,
 and set it and update it from the outside, like whenever a user changes the config colours,
 typically via the configChanged() slot.
Only units directly in the core (such as the components folder) can use global variables directly.
If your widget absolutely must use a global variable, then you must move it to the
 components folder, where all the units there had no easy choice but to use them.
Widgets folder: Completely independent. Components folder: Not-so-independent.

Another folder to watch out for: Anything in the libs folder. No global references there at all.

Finally, guys please try to remember to test builds with the MODULES_BUILD_STATIC
 cmake option set **AND** the following linker flags set:
-DCMAKE_SHARED_LINKER_FLAGS=-Wl,--no-undefined
(I leave them set all the time just to be absolutely sure.)
It will reveal any such dangling linker reference problems like this one.
This is crucial for module independence. It is crucial for some packagers of MusE.

I am anxious to build this as I see you managed to squeeze a slider thumb in there,
 and move the markings, something I was reluctant to to do to save mixer strip space.
The small corresponding increase in width seems agreeable from the pictures,
 but I'd like to see it up close. Looks great though! Thanks.

terminator356:
Built without the static modules and linker flags (it always will).

Wow, it even keeps the original slider markings behaviour in the standard light theme mode,
 neat trick! The thumb and markings should please a lot of users.
The look is pretty cool man. The knobs are reminiscent of a certain product, and of
 our Deicsonze MESS synth, which I imported from Werner's original MusE-2 (in git attic).

(I know it's fresh, but if I would suggest anything at all, possibly could I suggest slightly
 less primary track colours? Just a tinge of less primariness (?))

If you noticed some commented code in the meter unit, that was a fully working attempt
 to put the peak values **inside** the meters. It worked but the text obviously had to be vertical,
 and soon afterward someone added the actual peak boxes, so I disabled my code.
Also there's some code to draw markings inside the meters! But it needed improvements.
I am most concerned (paranoid) with speed in the meters. Tests showed heavy penalties
 for careless meter drawing with a few dozen, full height, meter instances in a big project.

Another thing about markings: I've been meaning forever to 'join' the slider, it's markings,
 and the meter together in one control (I've got some skeletal code in there somewhere),
 **or**, at least have the slider inform the meter, somehow, of the slider's current minimum
 and maximum marking position (the thumb limits), so that meter top and bottom match them.
It's a fairly important goal. The second idea is better, and modular.

Not sure if I ever mentioned, about 'joining' the control and meter classes:
At one point I was trying to make a very **compact** 'knob with meter surrounding it'.
Think of a knob with two 'C's 'cupping' it like [O] for stereo meters, and a single fuller 'C' for mono.
A real space saver on the strip, eh? Part of my continuing... ahem, 'compact' widget series.

kybos:
Thanks.
The build problem is solved (and I added the required flags to my debug configuration).

> Wow, it even keeps the original slider markings behaviour in the standard light theme mode,
> neat trick!

Yes, the position of the slider scale is also customizable now (a stylesheet property), it can be on the right or left, or over, or hidden. I noticed that most of the controls were originally copied into MusE from the qwt project, so they offer a lot of flexibility out-of-the-box.

This PR is mainly about providing the framework for extensive customizing of the UI. The new theme itself is rather a proof of concept for this, and a work in progress ;-). Sure the colours and other things can be improved. If I have time, I will perhaps create some more themes to show what is possible.

Tim, if you don't like the current (strong) colours, it's also easy to change them yourself (in Appearance->Colors). Your customized colours won't get lost any more when you change theme, they will be saved in the home config folder, and reloaded when you switch back to this theme.

kybos:
The issues I was aware of should be solved now, so I would be ready for a merge (I also desaturated or changed the track colours a little that Tim did not like ;-) ).
Could you please check if everything is working for you?

I would keep things quiet for a while before thinking of new release, that would give me a chance to catch potential glitches if any, and polish the new theme and perhaps create some more themes.

spamatica:
dark_flat_artefacts
Seems to work fine on my old trustful laptop but on another computer I happened to try it on this wierd thing happened with the top of the mixerstrip.
I tried removing the configuration and such but no nice.
It has kubuntu 18.04 just like the laptop. Has a nvidia graphics card though that sometimes gives wierd artifacts. (The fonts used to start really wierd but they are just fine now... maybe some of the hidpi fixes)

Apart from this wierd error which I'm not sure is "real"... I have one request.
When clicking a button, could you add some visual feedback that the button is clicked? I'd suggesting making it white while the button is held.

kybos:
> Seems to work fine on my old trustful laptop, but on another computer I tried it on this wierd thing happened with the top of the mixerstrip.
> I tried removing the configuration and such but no dice.
> It has kubuntu 18.04 just like the laptop. Has a nvidia graphics card though that sometimes gives wierd artifacts. (The fonts used to start really wierd but they are just fine now... maybe some of the hidpi fixes)

This is really strange (and not reproducible on my systems). The height of the label is even set to fixed size of 16 pixels in the new theme (to avoid the variable, font size dependent heights like in the other themes). The grid looks ok as well, the elements are put into separate rows and there is no row span or something.
 
> Apart from this wierd error which I'm not sure is "real"... I have one request.
> When clicking a button, could you add some visual feedback that the button is clicked? I'd suggesting making it white while the button is held.

I am not quite sure what you mean. I suppose the toolbar buttons? When such a button is pressed, it should get blue background - is it not working? Or do you want a distinction from the active "switch" buttons that are already blue?

terminator356:
@spamatica Can you tell @kybos what style your system uses,
 or which style you have chosen in Appearance, so that he might select it?

Here I am getting the artifacts as well. Here is standard system Oxygen style, 
followed by me choosing Adwaita style in Appearance.
Notice also that some of the text in places is cut off in Adwaita style
 especially in the toolbars, and other styles I chose do similar funky things:
mixer kybos 1
mixer kybos 2

spamatica:
Okay,

I think I originally had Oxygen-style but tried a few others and got the
same result so I didn't test that more.

Den sön 31 maj 2020 kl 23:43 skrev Tim <notifications@github.com>:

spamatica:
I am not quite sure what you mean. I suppose the toolbar buttons? When such
> a button is pressed, it should get blue background - is it not working? Or
> do you want a distinction from the active "switch" buttons that are already
> blue?
>

Sorry didn't reserach it enough to say for sure..  I think the toolbar
buttons are working.

What I mean affects dialog buttons at least. For instance [Ok] [Cancel]
buttons, I can't see them giving any indication that I pressed them.

kybos:
> What I mean affects dialog buttons at least. For instance [Ok] [Cancel] buttons, I can't see them giving any indication that I pressed them.

Thanks, now I see. Yes, this needs some more tweaking.

kybos:
> > What I mean affects dialog buttons at least. For instance [Ok] [Cancel] buttons, I can't see them giving any indication that I pressed them.
> 
> Thanks, now I see. Yes, this needs some more tweaking.

It should be OK now

kybos:
As for the KDE issues:

I was able to reproduce (and hopefully correct) it on a KDE system. I tried all the Qt styles in Appearance, I can't see any problems. I don't have Adwaita there, however, it does not seem to be a KDE standard style (I thought Adwaita was GTK style anyway...).

BTW: I also have many Qt styles for selection in Appearance on my Cinnamon system, but none of them works properly with MusE except Fusion. I think it's an impossible task to try to make it work with all existing combinations of desktops/styles (which often have their own issues as well)...

terminator356:
>  I think it's an impossible task

As I mentioned elsewhere, the OOM fork of MusE removed all appearance features,
 including colours, in favour of an all encompassing master stylesheet.
I tried to keep both and make them work together, back when styles were few
 and reliable. Now it's a losing battle. Silly styles like BB10, which is supposed
 to look like a Blackberry, completely mess up all widgets.
Although I do not want to remove custom colours or fonts, and should not have to, I suspect
 that slowly a master stylesheet is where we are headed whether we like it or not.
Gradually every type of widget is going to have to be customized in the stylesheet,
 so that styles do not mess with them. Does this sound correct?

terminator356:
Maybe until we have some sort of master stylesheet, one somewhat undesirable
 possibility is that we require Oxygen to be installed, and force it to be used always, at startup.
We could check the available styles at startup and drop an error or warning if it's not there.
Thus Oxygen's style sort of 'becomes our stylesheet' for the time being.

terminator356:
Would be cool if there was some automatic tool to generate a stylesheet
 from a given Qt style - all of the style's features would be replicated in the stylesheet.

kybos:
IMO it was the right decision to keep the appearance colours, they are very user friendly and fun to play with. I don't see a real conflict here, these colours are MusE internal and typically not accessible from stylesheets (I see no reason to expose them, that would only cause more confusion and problems).

One big stylesheet for everything else is one possible way, but huge amount of work. You have to put in virtually everything, every attribute of every widget, to be 100% sure (no idea if you can generate it from a style...).

Perhaps it's much simpler: I believe your suggestion to use Oxygen only goes in the right direction. It's pretty unusual for Qt applications to offer this kind of style selection. As I wrote above, from my 10 styles offered only one really works, so what's the use of it all?

But Oxygen is definitely the wrong one to keep. It's a KDE specific style and not found on most other Linux desktops (at least not as a Qt style). We should use a Qt's own built-in style to ensure compatibility over different desktops and even platforms. According to my earlier research on this topic, the most compatible, reliable and desktop/platform agnostic one is Fusion. I tested it also on KDE and the results are as expected. So what about setting that as default?

terminator356:
Whichever one works best, I suppose.
We'll have to warn if it's not present, and be sure to force it to be set once found.
I guess that means we would remove the style from Appearance settings :-0
I suppose nobody really fools around with that setting anyway unless trying
 to force it back to Oxygen or something.

terminator356:
Hm, one problem with this approach is that **everything** will be the chosen style,
 including title bar widgets, menus, and all manner of behaviour. Perhaps this would be
 annoying for some as they might expect that at least the basic look and feel matches
 their desktop style.

spamatica:
Den mån 1 juni 2020 kl 21:21 skrev Tim <notifications@github.com>:

> Hm, one problem with this approach is that *everything* will be the
> chosen style,
> including title bar widgets, menus, and all manner of behaviour. Perhaps
> this would be
> annoying for some as they might expect that at least the basic look and
> feel matches
> their desktop style.
>
I'm leaning towards trying to lock things down a bit, we can't all have
everything, we will never get anything done... Trying the styles I have
here, only a few of them look sensible, the rest look simply awful so
supporting them is just an illusion.

A start could be to not expose the style setting in the configuration
dialog but keep it in the configuration file, so people that want enough
rope to hang themselves have that option.

What is left though is deciding which style to select... which is also
where this might not work that well.
On my system Kubuntu 18.04, I have these styles installed:
- Breeze
- Oxygen
- QtCurve
- Windows
- Fusion
of which at least the two first are kde-specific, and those are sadly the
ones that look best.
I'm sure there are plenty of other styles that can be installed but it's
hard to rely on styles that might not be installed.
What style are you using kybos?

terminator356:
"Enough rope..." Lol !
Thanks for reminding of the style ownership thing, KDE vs Qt vs Gnome vs other.
Maybe warn about style installation "This application works best with ... style"
(And don't we all.)

kybos:
> What style are you using kybos?

I use Fusion, for the reasons given above.
I think the default setting after fresh installation was the Qt's GTK style (which makes sense), but it had many issues with MusE on Cinnamon. So I did some research and found the recommendation to use Fusion (especially on GTK based desktops), as the most compatible and universal one, the alternative being the Windows style, which is almost as good and also Qt built-in (but who wants Windows style on Linux...).

As mentioned above: Fusion is a built-in style, it's part of the Qt installation, so it's always there, no need to check its availability. 

If Fusion doesn't look good for you on KDE, then we can perhaps keep the current state, but at least put some recommendations/disclaimers in the wiki. 

@spamatica: Have my recent fixes solved the issues with the mixer labels on your system?

terminator356:
Just a heads up: I have fixed the blank mixers upon auto-restart and
am trying to get the last file opened to auto-reopen (it ain't easy).
Hang in there, watch for ChangeLog, relief on the way for all of this style testing and restarting...

spamatica:
@spamatica <https://github.com/spamatica>: Have my recent fixes solved the
> issues with the mixer labels on your system?
>
If you mean the top of the mixer, no :(
I realized that the big track label has been like that all along, something
is really weird about this particular installation... could there be some
qt configuration file that interferes?!
The feedback on clicking buttons looks fine though!

spamatica:
Den tors 4 juni 2020 kl 20:01 skrev Robert Jonsson <spamatica@gmail.com>:

> @spamatica <https://github.com/spamatica>: Have my recent fixes solved
> the issues with the mixer labels on your system?
> If you mean the top of the mixer, no :(
> I realized that the big track label has been like that all along,
> something is really weird about this particular installation... could there
> be some qt configuration file that interferes?!
> The feedback on clicking buttons looks fine though!
>

Hey! Looks like I wasn't paying attention... or trying the wrong thing, or
whatever... it is working now!

terminator356:
- Fixed blank mixers upon auto-restart.
- Reload the last song upon auto-restart. Templates included, and midi config loading choice.
Was tricky, possible wee corrections later...?

- Make MusE QMainWindow own the mixers. They had no owner, no deletion.
Leak? Test OK so far.

On that last fix, lemme know if any unusual sightings with the mixers, if anything changes.
Changing ownership might alter default properties passed down to the mixer widgets.

kybos:
> On that last fix, lemme know if any unusual sightings with the mixers, if anything changes.
> Changing ownership might alter default properties passed down to the mixer widgets.

Tested, no issues so far.

kybos:
> Hey! Looks like I wasn't paying attention... or trying the wrong thing, or whatever... it is working now!

Good to hear ;-).
So can we merge this now, no showstoppers?

We can continue the general Qt style handling discussion in the dev space, it's not directly related to this PR.

terminator356:
OK bud. You got it ;-)
There are too many changes to examine line by line, and they are mostly
 appearance related, so let's do this and move forward.
Thanks.

PR #813, opened 2020-05-24
added support for adjusting automation for structure cut/insert

spamatica:


spamatica:
Compiles, opens, no immediate problems atleast.
Though I can't see much difference. :P 
Browsing the changes it looks like there is some optional 3d-look. Or rather the new flat is probably the optional one. But I couldn't find where to control this?

terminator356:
Thanks for looking into this.
One more thing that was left 'dangling' that I forgot about.

spamatica:
No problem, I was a bit afraid that it would be terribly difficult since
you left it, but I think I managed :)
...It's good to have these bite-sized features to work on. Bigger tasks I
seldom get done..

Now, I just need to get back to that other pull request with hidden events.

Den sön 24 maj 2020 kl 21:03 skrev Tim <notifications@github.com>:

PR #814, opened 2020-05-26
New/changed Python scripts

StaffanMelin:
Added a SpeedHalf script and modified the DoubleSpeed script and changed its name to SpeedDouble.

The name change was to place the two scripts next to each other.

The change was to change not only the tempo but also the note length.


spamatica:
Seems fine, merging :)

PR #818, opened 2020-06-08
Fix buid with Qt >= 5.15

schnitzeltony:
There were several errors as
| error: aggregate 'QPainterPath path' has incomplete type and cannot be defined

Signed-off-by: Andreas Müller <schnitzeltony@gmail.com>

### DIFF
--- a/muse3/awl/knob.cpp
+++ b/muse3/awl/knob.cpp
@@ -25,6 +25,7 @@
 
 #include <QMouseEvent>
 #include <QPainter>
+#include <QPainterPath>
 
 namespace Awl {
 
--- a/muse3/awl/slider.cpp
+++ b/muse3/awl/slider.cpp
@@ -24,6 +24,7 @@
 
 #include <QMouseEvent>
 #include <QPainter>
+#include <QPainterPath>
 
 namespace Awl {
 
--- a/muse3/muse/components/slider.cpp
+++ b/muse3/muse/components/slider.cpp
@@ -26,6 +26,7 @@
 #include "mmath.h"
 
 #include <QPainter>
+#include <QPainterPath>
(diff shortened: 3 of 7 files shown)


terminator356:
Oh hi there !

Moments ago I pushed a whole whackload of fixes related to Qt 5.14 / 5.15 deprecations
 including all that QPainterPath stuff and much more.
Then I read this pr.

I upgraded my OS and got hit with this, and it took several hours to fix.

Your pr could probably be discarded now.
Let me know if any trouble with these fixes, there are many, and one of the
 areas to watch out for are the QT_VERSION checks that I do when I do an
`#if QT_VERSION >= 0x050***   #else   #endif`
That always takes time and research because we must get the 'switchover points' correct,
 in this case for all users of < 5.14, 5.14, and 5.15.

OT: I'd like to ask a question: Can you tell me if **any** of your DSSI synths
 such as Hexter, XSynth etc. are showing their native UIs in any DAW
 such as MusE, Rosegarden or QTractor ?

Thanks!

schnitzeltony:
Had a quick test with Qt5.14 / qtractor and all DSSI plugins on my Raspi:

hexter: no GUI opens / no fallback dialog opens and I am sure I have seen it long time ago!
Xsynth: GUI ok
FluidSynthDSSI: GUI ok
LinuxSampler: Fallback dialog opens
nekobee: GUI ok
WhySynth: GUI ok
YK Chorus: opens fallback dialog - have no idea if there is a GUI for DSSI
ZynAddSubFx: fallback dialog opens

terminator356:
OK thanks for the test. Hm, they don't work here after this upgrade.

Also, a small update fix has been applied in master to the previous fixes, see ChangeLog,
 be sure to get them all.

PR #820, opened 2020-06-10
Remove mixer strip toggle button

kybos:
We discussed that a lot already:
I removed the ugly button for hiding the mixer strip and cleaned up the layouts a bit.
The function is now accessible from menu (Display->Toggle Mixer Strip) or by pressing F3.
Please check so it doesn't come by surprise ;-).


terminator356:
Hi. I hate to nit-pick but I believe there is another reason why the button is there.
See the bottom area of the track list? It is empty in the screen shot.
When the track list is full there will be a track occupying that bottom empty space,
 but the horizontal scrollbar is blocking the view of its corresponding canvas track
 and any parts on it.

I think whoever put the button there did so not only to have the button, but to fill up
 that empty track list space and make things more even. Sort of wysiwyg.

I would like to try and go in the other direction, it seemed the logical way to go, 
 that we add the button to the editors as well.
It would be straightforward, we raise up the bottom of the mixer panel and put
 the button there. This would not interfere with the 'evenness' of the bottom area
 of the editors, unlike the Arranger where it would interfere if removed.

Also, there is a simple solution to the icon problem: Show a standard grey or black
 left arrow when the panel is open (meaning 'close'), and a right arrow when closed
 (meaning 'open'). Thus it apears sort of like a 'drop-down' combo box or list, except
 it open towards the right.
Whaddya think?

terminator356:
Sorry, I completely forgot to offer this solution for your desire to remove the button:
Just have to imagine your screenshot with a small correction:
We remove the button.
Extend the mixer panel all the way to the bottom.
Everything from the left edge of the track list, to the right edge of the app,
 remains the same - that is, the bottom area is the height of the scrollbar.

Sure, the bottom area of the track list will be grey, but look at it this way:
Probably better that than a track list row and the h-scroll is blocking the
 view of its actual canvas track.
We... could put some sort of track-related info/status text in that area...
Over in the drum editor, I have a need to show whether the displayed drum list comes
 from a drum map file or is built in to a synth. It helps users know what's going on there.
Bottom of drum list is a good ideal candidate place to show that text, not say, some
 grand app-wide bottom status bar.

terminator356:
Yap yap yap... Want to mention I saw the neatest thing:
Just tried Android Studio and found its (java style?) 'invisible' scroll bars quite striking.
There's **no** filled thumb 'track' background at all, and the thumb is grey but **transparent**.
I thought "All for an extra 20 or so pixels of content view...". Smart though.

kybos:
My thoughts:

1. I was aware that the track at the bottom is possibly not fully visible in the part canvas because of the scrollbar there. But I don't see it as a drawback, just the opposite, it's added value, you see more information on the tracks and can just scroll up a little if you want to see the parts. It seems somewhat illogical to me to intentionally hide that information.
 I understand your point, of course, but still I don't think that the symmetry is disturbed by the new layout (it was certainly more disturbed by that ugly button ;-) ). The track list and the part canvas are separate windows so it's nothing unusual that they have different scrollbar handling/display.

2. The width of the scrollbar is a variable and customizable value, so I don't think it's a good idea to put buttons or texts somewhere relying on that width. I see this problem on my system in the midi editors, with the "Add controller view" button, where the former text "Ctrl" was cut off at the top (it looks much better with the new icon now, but even the icon does not fit completely and overlaps with the button border).
It's possibly related to the fact that in the typical KDE styles the scrollbars are quite wide, and the MusE development has obviously been very KDE-centric. Given the fact that KDE is only one of many desktop environments for Linux (not even one with the highest "market share"), and it's also very special in many respects when it comes to the styles, IMO it's not a good starting point for widely compatible development (I'm off course not trying to put you guys off KDE ;-) - it has become quite nice and even fast after all the disasters of the past - I just wanted to mention that aspect too).

kybos:
@spamatica 
It would be nice to have your opinion too ;-).

kybos:
I added a scrollbar to the track list too, it's useful anyway and solves the problem with the bottom track assymetry (this is also how it's done in qTractor):

image

I think it's a good compromise and hope everybody is happy now ;-). Sure it looks better this way.

terminator356:
Freakin' brilliant! Actually much needed due to the very wide list.
I often find it annoying to have to expand that handle.

I agree with this.
But... back to the whole button thing:
I feel an expand/close button belongs there, somewhere close at hand.
As I proposed, use an unobtrusive 'stock' black or grey arrow-head instead of the green arrow,
 pointing left or right depending on close or open mode. Simple.
Where else would such a button belong? Any ideas?
That space down there literally begs for a button to be there.
It makes the whole bottom area even.
Then we can add it to the editors as well.

For example, take a program like Okular (pdf viewer).
It has the left-hand side panel with thumbnail and contents buttons etc.
Clicking those buttons opens or closes the chosen **expandable** info panel.
It conveniently remembers the info panel size so users don't have to
 keep dragging the handle every time.

Any opinion @spamatica ?

kybos:
There is perhaps a misunderstanding - I did not remove the functionality connected to the button (=hiding the strip), I even extended it, now it can be called either by pressing F3 (definitely much faster and user-friendly than pointing the mouse to the tiny button), and it's also available in the Display menu.
As I tried to explain above already, a button should never be dependent on the scrollbar width, because the scrollbar width is variable and can become very small with some styles.
I see no contradiction in implementing the strip hiding function in the midi/wave editors as well, following the same principle.

terminator356:
Aaawright. OK. Let's get this in. Thanks.

terminator356:
Artefact? Very intermittent. Either there or not there (flashing depending on mouse location).
track_list_new_scrollbar_artefacts_1

kybos:
I guess I will have to try with KDE, I had no such problems on my system, tested quite a lot.
Everything looks so different in your screenshot. We really will have to do something about the Qt styles, it's so unpredictable as it is now...
I will look into it.

terminator356:
Next time I boot, I'll try to go into Mint and check.

After many repeated reproduceable tests:

It happens only with Breeze so far, not Fusion or GTK. I don't have Oxygen at the moment.
On Breeze it happens with or without stylesheet (ie. light theme).
It seems to trigger by moving the mouse into the track list area.
Maybe coincidence, but Breeze has this bright blue 'hover rectangle' surrounding
 the track list. Maybe some extra style pixels not accounted for in the code?

kybos:
I tried but couldn't reproduce in Mint.
Wanted to test in KDE VM, but something has broken there, I can't start it.
I will continue tomorrow.

kybos:
I was now able to test it in KDE.
This issue only occurs in Breeze of all themes, not even Oxygen is affected.
So it seems to be a Breeze bug. I noticed several other issues with Breeze (and Oxygen too, BTW), so it will be necessary (at least for my new Dark flat style) to set Fusion as a base style if it's supposed to look as intended. 
The KDE styles (as C++ style plugins) don't work well in combination with additional stylesheets, they ignore many css directives so are not as complete and compatible as the Qt own styles like Fusion.

I suggest we continue the style discussion in the dev space.

PR #829, opened 2020-06-22
Pr fix wrong slider groove height

kybos:
After Tim's recent meter/slider scale alignment, the grooves were misaligned. This was more a slider bug, the groove was unrealistically long. Fixed now.

### DIFF
--- a/muse3/ChangeLog
+++ b/muse3/ChangeLog
@@ -1,3 +1,5 @@
+22.06.2010
+	- Fix slider/meter groove misalignment in mixer strip. (kybos)
 20.06.2010
     - Cmake option changes (Tim): Removed ancient ENABLE_VST and ENABLE_EXPERIMENTAL,
        and dangerous ENABLE_LV2_MAKE_PATH, and enabled ENABLE_PYTHON by default since
--- a/muse3/muse/components/slider.cpp
+++ b/muse3/muse/components/slider.cpp
@@ -69,16 +69,16 @@ namespace MusEGui {
 
 Slider::Slider(QWidget *parent, const char *name,
                Qt::Orientation orient,
-               ScalePos scalePos, 
-               int grooveWidth, 
-               QColor fillColor, 
+               ScalePos scalePos,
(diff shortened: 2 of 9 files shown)


terminator356:
Wow, thanks it never occurred to me.

I wonder if you could have a wee peek at the horizontal versions of the sliders,
 found in the generic plugin UIs. Just means including your recent changes in the
 horizontal sections of the code.

I too neglected that when I fixed the meters, a small routine that gets the margin,
 didn't check if it was horizontal or not. Although there's no harm since no
 horizontal slider has a meter associated with it yet.

The thumb changes and so on seem to have been applied only to
 the vertical code sections.

terminator356:
Couple of things: Why a clipper? To get a rounded shape? Could the painting path
 not be modified directly? 

I recommend changing the new enumerations in MusECore::Corner
 to more explicit names, by prefixing them with 'Corner' since for
 example MusECore::All is a little too vague and might interfere with
 some future enum (although that name would be discouraged).
Like MusECore::CornerAll etc.

kybos:
> I wonder if you could have a wee peek at the horizontal versions of the sliders,
> found in the generic plugin UIs.

It's on my list, sure we want to be consistent ;-). 

> Couple of things: Why a clipper? To get a rounded shape? Could the painting path
> not be modified directly?

I gave it a try, but it was more tricky than I liked (especially the bottom part of the groove), never got it pixel-exact and without side effects for all parameter variations I tested (unfilled thumb etc.).
So I thought a simple clip is clearer and more readable anyway, no intransparent calculations necessary. I think Qt is very efficient and highly optimized for such basic painting routines, but right, there is the additional calculation of the clipping rectangle - do you think it's so slow?
I can give it another try when there is time, or you can change it if you feel like it, you probably know better than me what all these values and positions mean and how they play together.

> I recommend changing the new enumerations in MusECore::Corner
> to more explicit names

You are right. I did not create the enum, however, it was already there, I only added the _All_ value. I fixed it in master.

PR #849, opened 2020-08-22
Pr new gui style

kybos:
As discussed in dev space...


terminator356:
Thanks. A new era, perhaps?

PR #867, opened 2020-10-10
Pr status bar

kybos:


terminator356:
Suggestion: You're welcome to somehow take the overlay text from the midi controller graphs
 and put it into the status bar somewhere. There are two lines of text that are shown there, found in **ctrlcanvas.cpp**:

"Drawing hint: Hold Ctrl to affect only existing events" (general message shown until something is actually drawn) and

"Make the current part's track match the selected drumlist entry" (a warning put there by Windfisch that 
 you are not looking at the current part and should select it with Alt Left/Right keys.)

In my last commit I fixed the 'smearing' bug effect on those texts when scrolling horizontally.

kybos:
Sure, will look into it.
There are of course many other things that should or could be included in the status tips, but I had to start somewhere ;-).
Next thing would be to convert all the existing and useful Whatsthis texts to status tips.
And remove the whole Whatsthis thing if you agree. It's quite outdated and hardly used in current software any more. It's a nuisance for the user to always have to click on the icon and then blindly search the UI for an existing whatsthis text.
If you look at Blender, Gimp, Inkscape etc., all of them use some kind of status tip now (that's where I got the inspiration, as I noticed that it can be extremely helpful in complex programs).

terminator356:
I did not realize until I installed Mint that there were no whats this features at all in some WMs like that one.
One problem is that I wrote fairly long multi-paragraph crucial explanations in several of them,
 so that they could be 'at hand' with a click. Lacking any context sensitive help, it was the best thing I could do.

kybos:
That's true, for long texts the status tip is not suitable.
So we will keep whatsthis for the time being. I even added a new item for it to the Help menu so the user has the chance to know that there is a practical shortcut for it (Shift+F1).
I suspect KDE/KWin is the only WM on Linux that has this additional support for it. It seems it's rather a Windows thing that was never very popular in Linux.

kybos:
> Suggestion: You're welcome to somehow take the overlay text from the midi controller graphs
> and put it into the status bar somewhere. There are two lines of text that are shown there, found in **ctrlcanvas.cpp**:
> 
> "Drawing hint: Hold Ctrl to affect only existing events" (general message shown until something is actually drawn) and

I added this hint to the standard status tip for the CtrlCanvas widget.

> "Make the current part's track match the selected drumlist entry" (a drum editor warning put there by Windfisch that
> you are not looking at the current part and should select it with Alt Left/Right keys or click the part area.)

I had really a hard time to understand and to reproduce the display of this warning. I had to change the default Group setting in the menu to None, then it made some sense. But the warning is still useless IMO, because it's obvious when the track is inactive - the events are displayed in grey. And if you click one of them or create a new one then the track is activated automatically! Perhaps this was not the case in the past, so the warning was necessary?
I suggest we delete this message.

spamatica:
If you feel it superfluous, or wrong even, I would personally not mind if
you removed it.

Den sön 11 okt. 2020 kl 13:38 skrev kybos <notifications@github.com>:

kybos:
Thanks, warning is deleted.

PR #898, opened 2020-11-12
Pr mstrip

kybos:


terminator356:
I'm not sure if this is finished yet but I have some concerns:

1) The icons should only light up when active. A real mixer has lights behind the buttons.
A lot of work went into making it that way so we could easily see at a glance what is active on the mixer.
When there are dozens of strips, there are so many buttons staring at you it is very difficult to see what is actually active.
So I made them light up only when active, to really **grab your attention** - "Hey, this is on".
But now, all the icons are lit up like a Christmas tree. It looks like everything is on.
That makes it quite hard to see what is really active. Especially the solo/proxy buttons.

2) Please restore the expansion feature. A lot of work went into it to ensure we could see
the full text on controls at a glance. I even put a button at the top left of the strip's title bar to quickly
expand and collapse the strip. The expansion was unobtrusive, did not hurt anything, and was intentionally useful.
Even **Ardour** has such a button to expand the strips so users can see (somewhat) more text.
Why even bother with a splitter/expander handle now that text is not expanded? I am not in favour of removing all that work.
Keep in mind for example the Aux controls have user-settable track names that may not fit into the default space.
No one wants to have to hover the mouse over all those controls just to see what the text is.

new_mixer_1

terminator356:
Couple more minor things:

Instr/Ctrl: All the items on the 'Instr' tab - except for the instrument - are true midi controllers.
All the items on the 'Ctrl' tab are **not** midi controllers, they are track properties. There are no controllers for them.
I suggest renaming that tab to 'Prop(s)' and redoing the tooltip text to "Instruments and controllers" and "Properties".

For strips that have no Monitor and/or Record buttons, including them as greyed out tends to clutter the mixer.
I'm not sure how you could cure that now, since things have been rearranged.
I do understand that it makes things evened-out, and simply putting a blank area would not look nice.
But... It worked out before. I suppose I can live with this arrangement as it's not too bad.

kybos:
Thanks for your feedback.

1. That's right, I was thinking about it too. I will change it.
2. I never removed the expansion, it was an unfortunate last-minute bug I introduced when testing something (audio strip only). Just the opposite, I put some work into making the expanded strip look sensible (=symmetric, like in Ardour). Fix is in master.
3. Inst/Ctrl: Changed in master.

kybos:
It should look similar to this now (2 strips are expanded as example):

image

PR #899, opened 2020-11-15
Restore content_rating in appdata

hfiguiere:
Regression from commit 6e15fcc

cc @kybos 

This is needed for flatpak (and standard).

### DIFF
--- a/muse3/packaging/org.musesequencer.Muse3.appdata.xml
+++ b/muse3/packaging/org.musesequencer.Muse3.appdata.xml
@@ -6,6 +6,7 @@
   <project_license>GPL-2.0+</project_license>
   <name>MusE</name>
   <summary>The open source sequencer</summary>
+  <content_rating type="oars-1.1" />
   <translation type="qt">MusE</translation>
   <description>
     <p>


kybos:
This is no regression, I had to remove this tag to be able to build the AppImage, as it gives an error:

```
AppStream upstream metadata found in usr/share/metainfo/org.musesequencer.Muse3.appdata.xml
Trying to validate AppStream information with the appstreamcli tool
In case of issues, please refer to https://github.com/ximion/appstream
W - org.musesequencer.Muse3.appdata.xml:org.musesequencer.Muse3:9
    Found empty 'content_rating' tag.
```

It's only a warning from the validator in fact, but the appimage people are generally very strict about everything. So I can start a discussion with them about why a warning aborts the whole process, but this will certainly need some time and might not be successful.

On the other side, the empty content_rating is meaningless indeed. The OARS ratings makes absolutely no sense for at least 90% of applications. According to the specs, its an OPTIONAL tag, so if it's defined as obligatory in Flatpak, it can be considered a bug there, or at least a non-compliance.

A quick fix would be to put some harmless content in there (even if nothing is really applicable), e.g.:

 ```
<content_rating type="oars-1.1">
    <content_attribute id="language-humor">mild</content_attribute>
  </content_rating>
```

hfiguiere:
it's not harmless, it is factual wrong. 

If you use the generator: https://hughsie.github.io/oars/generate.html

you get exactly that element. it missing fail validating on Flathub.

kybos:
As expected...
https://github.com/linuxdeploy/linuxdeploy-plugin-appimage/issues/19

The oars generator produces an invalid tag if you put no content in it (according to the AppStream validator).

I fixed it in master so it should work both with AppImage and FlatHub. Try it and close this if it works.

PR #907, opened 2020-11-28
Pr tracklist mouse handling

kybos:
Hi guys,

Please check this PR and let me know your thoughts.
It's about the track list which has a pretty inconsistent and confusing handling of the mouse events. You could even say that each of the columns has a completely different handling.
As usual, I am trying to make it more standard and user-friendly. As it's not 100% backwards compatible, a special care has to be taken. On the other side, as we are going for a new major version, it should be done now.

So in brief:
The list now behaves in a standard way for the most part: The left mouse click selects the list item (including the multi and range selection I had added earlier) in all columns, except for the four flags (input/rec/mute/solo) which keep their former behaviour (I think this is acceptable, as you can also have e.g. check boxes in a standard list that are toggled by left click). 
Context menus are triggered by right mouse click where relevant.
The output port column had 2 different menus for left/right click (which I've never got used to... ;-)). The standard context menu (right click) now shows the output port menu. To show the plugin GUIs you can now:
- Right click on the track type icon for the synth.
- Middle click on the output port column.
- Shift + right click on the output column.
- Use the keyboard shortcut that Robert implemented recently (default is Ctrl+U). This works for synth tracks only, which is correct IMO.

I've been using it that way for a couple of days and I like it much better than before, but that's of course subjective ;-). Especially opening the synth GUIs is now so quick and easy to find, right-clicking on the synth icon of the track ;-).


terminator356:
Good stuff. Yay. I hated that we could only select a select a track by left clicking on the name.

The only thing I could suggest is some natural embellishment and progression,
 by allowing double click on the icon column to open a synth gui (or native gui).

We used to be able to open a synth gui by (double?) clicking on the port column of the midi or drum track that drives it.
Maybe add that to the right-click menu of the icon column of midi and drum tracks (where it says midi/drum now),
 just like you now have for the synth tracks.

Keep in mind many folks may not have a middle button.

In fact, as I was mentioning recently, I'd like to try to unify the whole thing, somehow (?), by putting
 **all** available entries in the left-click context menu (except perhaps for the automation menu).

Thanks!

kybos:
Thanks. A very good suggestion to better exploit the double-click possibilities, it's still quicker than the context menu for often accessed functions. It now works on both midi and synth tracks for both the icon and output port columns. I adjusted the key shortcut action for the synth GUI to also work on both midi and synth tracks in the same way and unified the code.

I will do some more testing and also adjust all the status tips/tooltips/whatsthis texts to reflect the new situation.

terminator356:
Awright cool. Anything here is definitely an improvement, so do merge when you wish !
Thanks.
I should be committing my latest raster fixes sometime today...

spamatica:
Hi there!

Oh, I hadn't realized multiselect had been added, cool!

As for the new stuff, a couple of things.
First a cosmetic thing. I'm not sure I like the 'n/a' text on previously
empty fields. Maybe it's just that it is unfamiliar, I feel it gets
cluttered though..
Funnily enough, comparing this to the old I now think it looks weird that
the fields are empty too... so not sure what to make of that. Maybe it's
just as well to put n/a, unless someone else has a better idea.

Next I was gonna ask for a right click menu to open a soft synth, it used
to be on the Port. I see now that the Icon still has the gui right click
menu.
In relation to that, the menu that is now on port for Soft synths now shows
the port-list. I may be missing something but isn't this mostly a problem?
(I'm having slight deja vu about this menu, like I have removed it before
so I MAY be missing something :P ) As far as I understand, editing the out
port for a soft synth only makes it misbehave, it should be bound to the
soft synth.
My suggestion is to remove the port menu for soft synths and add the gui
open menu instead. Btw I did notice that double click opened the gui.

Good stuff!


Den sön 29 nov. 2020 kl 18:27 skrev Tim <notifications@github.com>:

kybos:
> First a cosmetic thing. I'm not sure I like the 'n/a' text on previously empty fields. Maybe it's just that it is unfamiliar, I feel it gets cluttered though.. Funnily enough, comparing this to the old I now think it looks weird that the fields are empty too... so not sure what to make of that. Maybe it's just as well to put n/a, unless someone else has a better idea. 

I know it looks a little cluttered but it also makes clear to the user that there is no functionality behind the corresponding field. But I don't insist on the n/a at all, it's ok for me to leave the fields empty or perhaps put a dash or similar in there. You guys decide ;-). 

>Next I was gonna ask for a right click menu to open a soft synth, it used to be on the Port. I see now that the Icon still has the gui right click menu. In relation to that, the menu that is now on port for Soft synths now shows the port-list. I may be missing something but isn't this mostly a problem? (I'm having slight deja vu about this menu, like I have removed it before so I MAY be missing something :P ) 
As far as I understand, editing the out port for a soft synth only makes it misbehave, it should be bound to the soft synth. My suggestion is to remove the port menu for soft synths and add the gui open menu instead. Btw I did notice that double click opened the gui.

The output port menu on synth tracks is not my addition, it was there before (on the left mouse click, though). I noticed myself that its behaviour with the synth tracks was very weird, it did not make any sense to me, but as I'm still far from being an expert on the internal of MusE, I just thought I was misunderstanding something once again ;-). 
So I agree with your suggestion and will change it accordingly.

spamatica:
Den mån 30 nov. 2020 kl 14:55 skrev kybos <notifications@github.com>:

> I know it looks a little cluttered but it also makes clear to the user
> that there is no functionality behind the corresponding field. But I don't
> insist on the n/a at all, it's ok for me to leave the fields empty or
> perhaps put a dash or similar in there. You guys decide ;-).


n/a grayed out?
I don't know. :P Might as well leave it as it is now, unless Tim has strong
opinions about it?


> The output port menu on synth tracks is not my addition, it was there
> before (on the left mouse click, though). I noticed myself that its
> behaviour with the synth tracks was very weird, it did not make any sense
> to me, but as I'm still far from being an expert on the internal of MusE, I
> just thought I was misunderstanding something once again ;-).
>

Yeah, I know it was like that before (o:

> So I agree with your suggestion and will change it accordingly.
>
Yeah, great! Unless Tim has some insight that we both missed.

terminator356:
The input midi port for the synth tracks was my addition.
It allows users to choose another input midi port for the synth, that's all.
Without it, there was no way to assign the synth to a different input midi port.
Yeah, I know, it may seem a little weird since for all other midi tracks the column chooses the **output** midi port.

spamatica:
Do you mean that it can be used to connect an input device directly to a synth without a midi track? 
If that is the case then, yes, it is confusing. And what exactly would you use it for where a midi track would not be involved?

spamatica:
If there is a good use case, one idea to make it less confusing is to append this dialog to the 'Input routing' dropdown.

terminator356:
No. A synth has a midi port. This simply allows them to choose that midi port from within the track list.
Just like going to the Midi/Synth Configuration Dialog and changing there.
Except that here it is more convenient.

kybos:
> No. A synth has a midi port. This simply allows them to choose that midi port from within the track list.
> Just like going to the Midi/Synth Configuration Dialog and changing there.
> Except that here it is more convenient.

I think I know what you mean - it's the port ID that is automatically assigned when a new synth track is created.

I am not sure why anybody should want to change the port ID for a synth, it doesn't seem to make much sense (and you can do it in the midi config dialog if you really want, where it's at least a little clearer what one is doing). 
The context menu really is confusing, also because it looks exactly like the midi track menu (it even says Output port) and it gives the impression that you can interconnect synths with it or similar things.

Probably I am missing something, but I was always wondering why these port IDs are editable at all. Are they not rather a technical entity that should be assigned as needed by the app, like e.g. the track ID? Why the predefined 200, is there a technical restriction (all of them always saved - mostly empty - to the .med files)?

terminator356:
The midi ports offer an extra indirect layer.
We used to have less than 100 I think, then someone (with a large studio) requested more, so I increased them.
Why the extra indirect layer?
To understand why, look at the Audio Input and Output tracks.
They also offer an extra indirect layer. If the user's audio hardware setup changes, they only need to alter
 that one Audio Input or Output track, **not** all tracks connected to them.
If we didn't have that extra indirect layer, and we were to **eliminate** Audio Input and Output tracks completely,
 then every time they wanted to change something about the hardware (Jack) ports they would need to change
 every audio track connected to those ports.

Hope that makes sense.
(Hey, I didn't make it that way, he he.)

I have a plan for a while now, one of two possible ideas:
1) To eliminate the 200 midi ports and instead just have devices with a port ID number
 that can be moved up or down. Essentially the same thing as a port number - without actual ports.
2) Eliminate the fixed 200 midi ports and instead, obviously, replace them with a dynamic port list - unlimited ports
 and only as few or many as is necessary for the user. Essentially they become the midi equivalent of
 Audio Input and Output tracks.
I think I pointed out recently that in Werner's original MusE-2 (now muse-evolution in our attic),
 that is what he did. Midi ports/devices became actual tracks in the track list that the user could add or remove at will.
Over the years I've considered doing that here. It's a tough question - we could actually go the other way
 and move Audio Input and Output tracks out of the track list and into some config dialog just like midi ports,
 if you see what I'm getting at there, "out of the way". But those tracks have automation graphs.
But Werner actually had automation graphs for his midi port/device tracks as well ! Just basic limited stuff
 like master volume sysex commands.

Either way would be a **massive** undertaking, so I have not even started yet...

terminator356:
I know the synth popup menu may be a bit confusing but you'll dig it once you see it does the same thing as
 the midi/synth config dialog.
But hey, that popup menu was readily available at the time, and I didn't want to have to create a new one.
Maybe we can create a special popup just for that purpose.

kybos:
Thanks for the insights, Tim. 
I do not doubt the general usefulness of the port IDs layer, I never did. But I still don't get why the port ID for soft synths should be changeable - they are not hardware, they are fixed tracks. IMO it's more than sufficient that it can be done in the midi config dialog. Why confuse the user with this misleading context menu?
You know, it's not documented (AFAICS), both I and Robert failed miserably to get the sense of it, but you expect everybody else to grasp it? I'm a little sceptical ;-).

terminator356:
In your branch, in the context menu for the synth port column, above "Show GUI", can you find a way to 
 put "Add Midi Ports/Soft Synths..." and open the config dialog?

Actually that is one reason I recommend renaming that phrase to "Configure Midi Ports/Soft Synths..."
 because in this case you're configuring, not adding.

"Configure" covers all bases - adding **and removing** and configuring.

In addition, if you have time, I was going to attempt this but I don't want to interfere with this work:
When the configuration dialog opens, I would like to help the user by highlighting and scrolling
 the three panes to the device(s) and port(s) and soft synth(s) that **led to** the dialog opening.

For example if the dialog opened from a synth port column context menu, let's highlight and scroll to that synth
 in the device list, and highlight and scroll to that synth in the synth list (shows useful information about the synth),
 and... well, find a way to highlight that port in the port list, but unfortunately we don't highlight lines there
 and it would be tricky because when you click there a menu will open, however, I see no harm in highlighting the line.
If the dialog opened from clicking a Midi Track's output port or output routes context menu, let's highlight and scroll
 to the appropriate items in the panes.
Theoretically, we could even do this for the Midi Track's input routes context menu (multiple input routes are allowed).
I already allow multiple selections in the dialog's device list. Hm, now can we do that for the synth list and port list? Tricky.

Check out how I did something similar for the Instrument Editor - it opens to the appropriate current instrument
 of whatever opened it: A midi controller graph panel or a midi strip (but not the Arranger 'Midi' menu entry).

kybos:
> In your branch, in the context menu for the synth port column, above "Show GUI", can you find a way to
> put "Add Midi Ports/Soft Synths..." and open the config dialog?

Sounds like a good compromise. I will check.

> Actually that is one reason I recommend renaming that phrase to "Configure Midi Ports/Soft Synths..."
> because in this case you're configuring, not adding.
> 
> "Configure" covers all bases - adding **and removing** and configuring.

I absolutely agree. "Add" was not final, I thought about Add/Edit or similar, but Configure is better.

> In addition, if you have time, I was going to attempt this but I don't want to interfere with this work:
> When the configuration dialog opens, I would like to help the user by highlighting and scrolling
> the three panes to the device(s) and port(s) and soft synth(s) that **led to** the dialog opening.

I will look into it. I must also check the theming, as the list highlighting is currently too "heavy" on the default theme (pure white).

terminator356:
I had not noticed that the reused popup said "output port" so I see the reason for the confusion.
After consideration, that is the only place in the arranger where one could modify the midi configuration,
 which made it an oddball, so yeah, let's leave that popup out and try to direct users to the config dialog instead.
Thanks.

kybos:
I will merge this now, as I already had a merge conflict with master...
Essentially everything should be finished, some nice-to-have things kept for later.

PR #913, opened 2020-12-12
Fix fluidsynth version issues

kybos:


terminator356:
Neat stuff. I think I fixed some things when 2 came out but was unaware of this other stuff.
Thanks.

PR #937, opened 2020-12-30
Track list: Change track heigth for all/selected tracks

kybos:


terminator356:
Just want to add: If you've been thinking about adding a vertical magnification
 scroll bar to the arranger (ought to finally be done!), I have also thought about it a few times.
I concluded that it is important to use the vertical scroll value together with the user's adjusted heights
 in such a way as to **not** destroy the user's exact height values. That is, we cannot simply directly adjust 
 existing heights with the mag value as a factor. After a few movements of the scroll bar back and forth,
 the user's exact height values would lose accuracy, there would be loss.
Therefore we should (somehow?) apply the mag value as an additional factor of the track height value 
 stored in each track instance - **without** actually adjusting any of the user's track heights. Make sense?

kybos:
Yes, that makes sense and I am aware of it.
An independent view zoom would be nice-to-have, but I personally never felt the need for it. 
The arranger view is different from the other editors, just because the tracks have their own height attributes (which definitely makes sense). So I tried the exploit this very capacity as well as possible in this PR. 
I'm not even sure if a view zoom on top of it would be such a good idea from the user perspective, as in the end you would never know what the real height of the track is.
So, seeing that there are so many other open issues, some quite critical and important, I would give this a very low priority ;-).

terminator356:
Er, well, actually I just got a display with higher resolution (not hidpi) and now I'm wishing for this.
All my existing songs' tracks are too small. We have track height in Settings, but it only applies to new tracks.

> I'm not even sure if a view zoom on top of it would be such a good idea from the user perspective, as in the end you would never know what the real height of the track is.

Maybe we could make it work. Users would be free to adjust the heights at whatever mag factor,
 and the code calculates the correct value to be stored, such that readjusting the mag back to 1/1 would make it
 appear as 'normal' although we would clip the height to the minimum value in Settings track height.
Something like that I suppose...

kybos:
> All my existing songs' tracks are too small. We have track height in Settings, but it only applies to new tracks.

So the new Ctrl+resize from this PR could be a quick help for the time being...

PR #966, opened 2021-02-13
Saveas rework

spamatica:


kybos:
Looks good! 
Only the key mnemonic for _Save as new project_ should be changed, it's the same as for _Save as_ now.

Another idea: What do you think about adding another entry for _Save as template..._? This could help users to clearly see that there is such a function at all - currently it's only possible from the Project dlg, and there it's not so apparent. And you even don't have to use the Project dlg (there is this global setting).

spamatica:
Oh, woops, forgot about the key name. I will fix it.
Good idea with template too. I will check it out.

PR #1038, opened 2021-07-12
Prevents heap usage after free

7u83:


terminator356:
Thanks! Good catches.

Issue #1037, opened 2021-07-09
[Feature Request] Time stretching/shrinking for midi/drum parts


cezaryece:
Sorry if it is possible right now and I dont know how to do it, but if it not possible I vote for that possibility.

If it is possible just tell me how I can adjust duration of some midi part.
Let say I have 2 bars with 4 quarter-notes each (2 x 4 = 8) and I would like to get 3 bars with same 8 quarter-notes shifted in time to fill 3 bars.

spamatica:
There is currently no way to stretch midi dynamically.
Our scripting interface could be used to make such a stretch tool. Currently there are the related scripts doublespeed and halfspeed.

Issue #1022, opened 2021-06-02
Allow even more zoom out in piano roll


StaffanMelin:
I would like to be able to zoom out even more, like 1 "grid box" per bar.

I work on some space music right now and using really long bass drone notes.

See attachment of current maxed zoom out.

![muse_zoomout](https://user-images.githubusercontent.com/33361311/120434150-61da7580-c37c-11eb-8ce0-f596c9ca4483.png)

cezaryece:
Vote for this!

PR #1065, opened 2022-04-04
Added script for MIDI timestretch and fixed small bug when receiving results

StaffanMelin:
Hi! Trying to get into MusE development, so I started with adding the requested script.

There was a bug in scripts.cpp, so when reading back the modified MIDI data from the script, all CONTROLLER events were misinitialized (tick was not read and set). I have fixed this too.

Re: https://github.com/muse-sequencer/muse/issues/1037


spamatica:
I didn't try running the script but it looks good. Good catch on the bug!
One last thing before merging, please add a line in the ChangeLog file along the same lines as the previous entries.

spamatica:
Regarding the related ticket. 
It is good to mention the ticket number (in this case #1037) in the related commit message, this way github will automatically reference the commit in the ticket.

StaffanMelin:
Sorry, somehow I messed up (still new to Git). Was trying to do another pull request fixing issue #1022, but somehow this got merged into the first pull request. So the changes in pianoroll.cpp are for the second fix (ie #1022).

Will this still be ok or do I have to redo it somehow?

spamatica:
Hey hey, sorry, missed the continuation of this. 
I see the other fix still there, looks trivial though so I'll go ahead and merge

PR #1070, opened 2022-05-04
Add keywords to the desktop file

snd1:


spamatica:
I'm not versed in how these are used but I guess they are just text strings, maybe add vst also?

snd1:
done :-)

spamatica:
thanks, added :)

PR #1077, opened 2022-07-12
Link with libatomic if needed for the architecture

ileanadumitrescu95:
Muse fails to build for riscv64 because it requires linking with libatomic for this specific architecture. This will fix the issue with riscv64 and any other architecture that has this libatomic link issue. Please note that building muse with these changes will require the extra-cmake-modules package to be installed first (I updated the README file accordingly).

### DIFF
--- a/src/muse/CMakeLists.txt
+++ b/src/muse/CMakeLists.txt
@@ -357,6 +357,10 @@ if(LV2_SUPPORT)
             OUTPUT_NAME muse_lv2host_module
             )
             
+      if(NOT HAVE_CXX_ATOMICS_WITHOUT_LIB)
+            target_link_libraries(lv2host_module atomic)
+      endif()
+
       target_link_libraries(lv2host_module ${LILV_LIBRARIES})
       target_link_libraries(lv2host_module ${SORD_LIBRARIES})
       target_link_libraries(lv2host_module ${LV2_LIBRARIES})
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -532,6 +532,10 @@ endif (ENABLE_VST_VESTIGE)
 ##
 
 if (ENABLE_LV2)
+find_package(ECM REQUIRED NO_MODULE)
+set(CMAKE_MODULE_PATH ${ECM_MODULE_PATH})
+include(CheckAtomic)
+
 set (muse_sord_ver "0.16.0")
 set (muse_lv2_ver "1.18.0")
 set (muse_lilv_ver "0.24.0" )
--- a/src/README
+++ b/src/README
@@ -76,6 +76,9 @@ details.
 
       - LADSPA (development file ladspa.h)
          www.ladspa.org
+      
+      - extra-cmake-modules >= 5.94.0
+         http://invent.kde.org/frameworks/extra-cmake-modules
 
          
   =============================


src/muse/CMakeLists.txt
```diff
@@ -357,6 +357,10 @@ if(LV2_SUPPORT)
             OUTPUT_NAME muse_lv2host_module
             )
             
+      if(NOT HAVE_CXX_ATOMICS_WITHOUT_LIB)
```
terminator356, line 360: Hi, thank you for the patch. I am ready to merge, but I am curious...
Is atomic only required for our LV2 support module?
If so, would it be possible (and wise?) to move the all of additions in the top cmake file
  down a bit further and place them inside the 'if (ENABLE_LV2)' block near line 505?
Our LV2 support is optional, that is why I am asking.

Maybe something like this?
Line 505: if (ENABLE_LV2)
find_package(ECM REQUIRED NO_MODULE)
set(CMAKE_MODULE_PATH ${ECM_MODULE_PATH})
include(CheckAtomic)

set (muse_sord_ver "0.16.0")
...

ileanadumitrescu95: Hi, I successfully tested your suggested changes and updated the commits. Sounds like a good idea. Thank you!

terminator356:
Good stuff, thanks very much!

terminator356:
Hi. I was unsure of the nature of the changes.
Upon full review I see what's going on now. I have made some changes:
I was the one who started using std::atomic some years ago.
It is used in a few places, including my "lock-free ring buffer" header files which are in turn
 used by our LV2 support  and several other places even when LV2 support is disabled.
Since our LV2 support is a module which is linked with our 'core' module, this explains why
 your fix worked for you - libatomic linked to LV2 module which is linked with core module.
Dependencies were thus resolved.

So... I have moved your top cmake file code back to the top of the file so it always runs.
And in the muse folder's cmake file, I linked libatomic with our **core** module instead of our LV2 module.

Sound good? Please pull my changes and try a fresh build if you can.
Let me know if any trouble.

Side note: I am struggling with our git automatic actions (our AppImage builds).
After some hours I realized that the distro we are using (Ubuntu 18.04) **does not** 
have CheckAtomic in its extra-cmake-modules package!
@spamatica  We must upgrade the distro to say, 22.04 LTS since 20.04 LTS 
does not have CheckAtomic either...
AppImage builds are **borked** until this is resolved.

terminator356:
> Side note: I am struggling with our git automatic actions (our AppImage builds).
> We must upgrade the distro to say, 22.04 LTS 

To avoid seriously major problems upgrading to 22.04, I solved the missing CheckAtomic
 problem by installing a local copy of CheckAtomic.cmake in our cmake folder.
In our top-level cmake file, if the CheckAtomic module is **not** found, 
we use the local copy instead.

Now our AppImage has finally completed building !
Please re-check with these new changes, if you can.

During CMake configuration, if you see this message:
"CheckAtomic module not found! Using local copy instead."
then you know it did **not** work correctly.
If you do **not** see the message, then everything went OK
 and it used your system's CheckAtomic module.

Thanks.
Tim.

ileanadumitrescu95:
I successfully built with riscv64 with your latest build. Thank you!

terminator356:
OK Thanks. One final change: 
I read that the CheckAtomic scripts will **automatically** make libatomic a requirement, if necessary.
That is good since it means we don't have to look for it in our cmake scripts.
Still, I updated the README to inform users that libatomic **might** be required for certain architectures.

PR #1080, opened 2022-07-26
Replace std::random_shuffle for C++17 compliance.

0EVSG:
The std::random_shuffle algorithm has been removed in C++17, see:

https://en.cppreference.com/w/cpp/algorithm/random_shuffle

Build fails with compilers that actually enforce C++17 compliance, particularly on FreeBSD / Clang.
Replace this code with the newer std::shuffle algorithm.

I did some brief testing and the random "Did you know" dialog seems to work.

### DIFF
--- a/src/muse/app.cpp
+++ b/src/muse/app.cpp
@@ -49,6 +49,7 @@
 #include <iostream>
 #include <algorithm>
 //#include <typeinfo>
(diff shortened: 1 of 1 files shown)


terminator356:
Thanks for the patch!

PR #1083, opened 2022-09-07
Fix mismatch new delete error

donarturo11:
Compiling from original branch from other directory than $srcdir/build causes mismatch new delete error. Adding [] to delete resolves this problem.


terminator356:
Good catch, thank you!

Issue #1084, opened 2022-09-08
Crash when trying delete complicated track


donarturo11:
**Describe the bug**
Trying delete more complicated track finishes with crash

**To Reproduce**
1. Go to selected track (see attached picture)
2. Right click to context menu
3. Choose 'Delete Track'
4. Program crashes: 
5. See error:
```
JACK: port connections changed: A:30 B:13 isConnect:1
JACK: port connections changed: A:31 B:14 isConnect:1
MusE: STARTING Heartbeat timer
MusE::seqStart: getting audio driver MusEGlobal::realTimePriority:5
[New Thread 0x7fffda8a16c0 (LWP 4714)]
graph_callback()
Thread <Prefetch, id 0x7fffda8a16c0> has SCHED_OTHER priority 0
unknown kbAccel 0x1000020
ACTIVE TOPWIN CHANGED to '<None>' ((nil))
ACTIVE TOPWIN CHANGED to 'Arranger' (0x5555558035b0)
ACTIVE TOPWIN CHANGED to '<None>' ((nil))
ACTIVE TOPWIN CHANGED to 'Arranger' (0x5555558035b0)
ACTIVE TOPWIN CHANGED to '<None>' ((nil))
ACTIVE TOPWIN CHANGED to 'Arranger' (0x5555558035b0)
double free or corruption (fasttop)

Thread 9 "muse4" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fffda9226c0 (LWP 4711)]
0x00007ffff36a14dc in ?? () from /usr/lib/libc.so.6
(gdb) bt full
#0  0x00007ffff36a14dc in  () at /usr/lib/libc.so.6
#1  0x00007ffff3651998 in raise () at /usr/lib/libc.so.6
#2  0x00007ffff363b53d in abort () at /usr/lib/libc.so.6
#3  0x00007ffff369567e in  () at /usr/lib/libc.so.6
#4  0x00007ffff36ab26c in  () at /usr/lib/libc.so.6
#5  0x00007ffff36ad042 in  () at /usr/lib/libc.so.6
#6  0x00007ffff36afa33 in free () at /usr/lib/libc.so.6
[...]

PR #1085, opened 2022-09-08
Implement checking if erase of element from list is possible

donarturo11:
Fixes issue #1084 .


donarturo11:
@terminator356 , could you review this commit, please? Thanks to this one, this program is more stable. It prevents of crashes, when object won't found in list.

terminator356:
Hi, thanks for the patch.
Unfortunately this is not the ideal solution, and is only a 'band-aid' solution.

We really need to know who is attempting to execute the operation with an invalid iterator,
 and correct the problem there.
To do this, I must ask, can you **post** (a link to) the actual song involved, or say a 'cut-down' 
 version of it that still causes the problem, or else a 'mock-up' that will cause the problem?
That way I can examine every detail such as song storage and so on.

Question: The song causing the problem: Is it an old song made before 2020?
Or is it a new song made recently?
If it is an old song, you may indeed be seeing bug 887 as you originally posted... Maybe...

Thanks.
Tim.

donarturo11:
> Question: The song causing the problem: Is it an old song made before 2020?

This song was created in summer 2022.

donarturo11:
> To do this, I must ask, can you **post** (a link to) the actual song involved,

[muzyka_na_smyczki_noSynth.med](https://www.dropbox.com/s/xldutw1hwa6joux/muzyka_na_smyczki_noSynth.med?dl=0)

donarturo11:
> Unfortunately this is not the ideal solution, and is only a 'band-aid' solution.

I tested another solution:
```
if(_imcv == _mcvl->begin())
          _mcvl->erase(_imcv);
```
Too runs. I thought about submit but I don't know if _imcv can be on another place of _mcvl. 
I wrote a simple text function to check, when _imcv is present in _mcvl and I noticed that when _mcvl tries to erase not existing item, program crashes. So I think that good way is to check if item is in list, then if condition is true, erase.

donarturo11:
Which better way to check if item is valid do you propose. In my humble opinion, direct erasing shouldn't be leave, because original solution is not as safe.

donarturo11:
I found a problematic block. Inside original song, trying to remove crashes program, but this block imported to new files, crashes after moving.
[test1.mpt](https://www.dropbox.com/s/iunpc6i3suwlo3z/test1.mpt?dl=0)

donarturo11:
@terminator356 I removed loop and used method ```void delMCtlVal(unsigned int tick, Part* part, int val/* = -1*/);``` .
Now this line looks: ```_mcvl->delMCtlVal(_imcv->first, _imcv->second.part, _imcv->second.val);``` .
Runs well. In my opinion it should be a a better solution.

donarturo11:
I think yet about reimplement the method erase inside MidiCtrlValue. It would be as override.

donarturo11:
@terminator356 I found yet another solution. I reverted ```_mcvl->erase(_imcl)``` in line 1282 of operation.cpp and uncommented line 35 of midictrl.h ```#define _MIDI_CTRL_METHODS_DEBUG_``` and the program didn't crash while delete track. 
I would like to ask if is any reason of optionally turn on this definition, and it won't better to remove ```#ifdef _MIDI_CTRL_METHODS_DEBUG_``` conditional?

donarturo11:
@terminator356 So I removed the commit with searching loop and replaced erase. Could you review the code after changes, please? If do you want to leave erase method in operations.cpp, maybe better enable debug methods as default in midictrl?
I shared entire directory on [Dropbox](https://www.dropbox.com/sh/g2bz4oed09asdxb/AACbLsqAvh-1SvHqq3TfklWsa?dl=0)

donarturo11:
@terminator356 Well, what you think about latest commit: [Use another method to erase item](https://github.com/muse-sequencer/muse/pull/1085/commits/38054144b68664342233990ee9fca48557e1526d)? Now it is good to merge?

donarturo11:
@terminator356 Well, what you think about latest commit: [Use another method to erase item](https://github.com/muse-sequencer/muse/pull/1085/commits/bbbf95ee6819f2523d37cb059768313cc5d455e9)? Now it is good to merge?

terminator356:
Hi, sorry for the delay, I am studying the matter and should have an answer this weekend.

donarturo11:
@terminator356 Sorry if I'm asking here. Do contributors use any platform like Discord or IRC? I know that comments below issues or pull requests are not good place for such discussions. A very good example is [Developer group chat](https://musescore.org/pl/handbook/developers-handbook/references/developer-group-chats) in MuseScore project.

terminator356:
Hi @donarturo11 
I have found several problems in our code. I am working to fix them.
I also found some bugs that, when fixed, should fix some features that were broken for a long time.
Please wait... Thanks.

luzpaz:
Any progress here ?

PR #1086, opened 2022-09-16
Fix shortcut hints in PartCanvas and tweak Normalize shortcut

theGreatWhiteShark:
- Instead of using hardcoded shortcuts in the item popup of `PartCanvas` the ones configurable in the settings are now retrieved and displayed. (For default settings the shortcut of "Normalize" did already diverge)
- The current default key combination of "Normalize" is shadowed by the "Fu&nctions" menu in the main menu bar. Changing it from Alt+N to Alt+Shift+N seems minimal inversive and make the action accessible again.


terminator356:
Sorry for the delay, thanks!

PR #1089, opened 2022-10-06
Add PI divisions

donarturo11:
This commit is to avoid MinGW build error: "M_PI_4 [or similar] is not defined".

### DIFF
--- a/src/muse/muse_math.h
+++ b/src/muse/muse_math.h
@@ -34,6 +34,22 @@
 #define M_PI 3.14159265358979323846
 #endif
 
+#ifndef M_PI_2
+#define M_PI_2 M_PI/2
+#endif
+
+#ifndef M_PI_4
+#define M_PI_4 M_PI/4
+#endif
+
+#ifndef M_1_PI
+#define M_1_PI 1/M_PI
+#endif
+
+#ifndef M_2_PI
+#define M_2_PI 2/M_PI
+#endif
+
 #ifndef HAVE_EXP10
 #define exp10(x) (pow(10.0, x))
 #endif


terminator356:
Thanks again for the patch!

donarturo11:
@terminator356 BTW I'm going to prepare to compile for Windows. At last MusE should be available for Windows users. Can we talk about it on discord, IRC or similar platform?

terminator356:
Hm, I tried to give you access to our githib member discussions, 
 but it requires giving full access to the repo.
We don't have any other discussion medium, other than the LM forums.
Maybe just for now the best thing is to open an issue and talk about it there.
That would be OK (there are other similar posts there).

donarturo11:
@terminator356 Do you know, why int64_t in MinGW is seen as long long int, when on Linux as long int? I noticed it while compiling on MinGW, where debug functions consisting %ld (eg. MuseFrame_t frame) throw compile error. 
Yet I didn't compile muse to finish, but I'm very near. 
As next step I'm going to write CI to automated compilation for this platform. I'm looking forward version for Windows, so I decided to take matters into my own hands.
I would like to compile it for Windows, because I would like to promote MusE as a good alternative for Cubase or Sonar, like I promote MuseScore as good alternative for Sibelius. As composer I use both MusE and MuseScore.

terminator356:
> Do you know, why int64_t in MinGW is seen as long long int, when on Linux as long int? 

I think that may be because on Windows a long int is still 32 bits. I can't quite remember,
 do a search, because yes there are differences in how Windows looks at long and long long etc.

I must warn, you might have some difficulty doing MusE on MingW.
I really hope you succeed, but from what I experienced with MingW and what I read,
 it does not support Linux threads, I seem to recall. I think that one must use 
 something like CodeWeavers products to get the full support needed for MusE.

I hope I am wrong but, I tried once to make MusE under MingW and it did not go well.
So my approach after that was to try to make MusE build natively under VC, 
 without MingW or other software. Obviously this is much harder, and I gave up after
 compiling about 80%, due to things like threads.

donarturo11:
@terminator356 I noticed yet, that if I'm trying to build some libraries as shared, I get undefined reference. In some modules CMakeLists.txt has libraries built always as SHARED. In Windows not better to give possibility to build as STATIC? Static linking, because of lack of RPATH mechanism, is often preferred.

spamatica:
> it does not support Linux threads, I seem to recall. 

Hey there, rare uncloaking... Hopefully more time and energy when winter comes on.

Regarding threads, I believe I switched out all pthreads for QThread in the platform_independence branch. Many years ago...
I didn't have the stamina to get through it all, but I believe it would be possible to do this maybe with some special cases for enhanced realtime-ness in Linux.

terminator356:
I did extensive research a long time ago on QThread to see if could do realtime.
I found it could not, even when set to QThread::TimeCriticalPriority.
I waded through Qt's source code to reach that conclusion. 

However, if I recall there are only two places where we need such realtime high thread priority:
1) Our ALSA midi driver loop
2) Our Dummy Audio driver loop

On Windows we might be able to completely forget about 1) since we would either be using Jack,
 or say the RtMidi library which might take care of any realtime threads for us.

For 2) it's a bit different. We would like to ensure that our dummy audio thread loop is not
 interrupted and is able to complete its loop tasks.
Although, I did take some steps to ensure that even when the thread is not realtime,
 it is still relatively safe from interruptions. Tests showed that it was mostly stable,
 but who knows what would would really happen when the thread becomes super busy
 with audio processing and some interruption comes along.
We could get rid of the dummy audio driver and stipulate that Jack or RtAudio driver 
 must be running. 
RtAudio **does** take care of realtime priority for us. 
(I contributed to the RtAudio project realtime code. I can't remember if Windows supports realtime though)
Apart from that there may be Windows realtime thread functions we could use directly.

So yes, maybe it's do-able, eh?

Issue #1098, opened 2022-10-19
Behaviour of editing views.


donarturo11:
Now, when I call list event, window is in dock, but if I call Pianoroll, normaly is opened in tab. I noticed if I call open Pianoroll in new window, until I won't change in preferences, has the same behaviour (open in tab).
What is expected behaviour. In my opinion, Event List should have the same behaviour as Pianoroll. About Pianoroll, if I have setted open Pianoroll in tab, default behaviour should be open in tab and calling open in new window should open in new window.

I propose to set Event List behaviour same as Pianoroll – open in tab or new window, not in dock, because is not so practical and comfortable (good solution was in MusE 3).
About Pianoroll, I think, that not opening in new window, if I have setted open in tab or opening always in new window if unsetted in my opinion is an unexpected behaviour.

terminator356:
Yes, opening the Event List Editor in a tab instead of a dock sounds reasonable.
But I do not know how to change this behaviour.
The developer @kybos wrote all of the dock and tab code, and it is all quite new.
I would need to examine the code but I am quite busy at the moment. I will try to look at it later.
But you are welcome to try to fix it...

donarturo11:
@terminator356 Thank you very much. So I'm going to work on Event List editor.

donarturo11:
> But I do not know how to change this behaviour.
But I'll find it.

PR #1099, opened 2022-10-26
Move List Editor from dock to window

donarturo11:
Resolves ListEditor mentioned in #1098 issue.

### DIFF
--- a/src/muse/app.cpp
+++ b/src/muse/app.cpp
@@ -2362,7 +2362,7 @@ void MusE::startEditor(MusECore::PartList* pl, int type)
       {
       switch (type) {
             case 0: startPianoroll(pl, true); break;
-            case 1: startListEditor(pl); break;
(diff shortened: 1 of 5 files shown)


terminator356:
Nice work. You are brave to wander deep into the guts of the source code ;-)
I suppose it is thankful that @kybos left some original lines commented.

I will pull and test later...

donarturo11:
> Nice work. You are brave to wander deep into the guts of the source code ;-) I suppose it is thankful that @kybos left some original lines commented.

I looked too into code of MusE 3.1.1.

terminator356:
Hi. I have reverted this pull. Please see ChangeLog for reasons why.

The code was kept. It can be switched on or off with a define in cobject.h:
MOVE_LISTEDIT_FROM_DOCK_TO_WINDOW_PULL1099
just in case we decide to move forward with that pull.

PR #1230, opened 2023-07-27
Debug substitutions

donarturo11:
In Windows some printf functions are compiled with warning, which in debug type is treated as error (default setting). Warning are caused by small differences in types like long int. To force reading as long int (%ld) I put casts for values.


terminator356 (commented):
Hi. The std::move was used for speed and efficiency.
Are you sure there is no std::move or equivalent on your system?
No Qt equivalent?
If no, then we should rewrite the function to accept a 
 pointer-to or a reference-to a QGradient, and fill it directly instead of
 returning a QGradient.
Tell me if there is no suitable replacement for std::move on your system.
If no, I will accept the PR but I will make the required changes.
Thanks.

donarturo11:
I reverted utils.cpp. Compiling error is only in debug mode, because return-type warning (in debug all warnings are treated as errors, maybe better if warnings like unused variables or return type wouldn't be treated as errors).

PR #1240, opened 2023-08-18
Allow building MusE with LV2 and recent kxstudio version

spamatica:
This pull-request is not for immediate consumption as I'm not sure it is universal.
But for lv2 package currently in kxstudio: lv2-dev/focal,now 6:1.18.10-1kxstudio2
The paths to headers under lv2/lv2plug.in no longer exist. 
The paths do exist under lv2/ though, so this pull contains these changes.


terminator356:
Thanks.
The LV2 include changes occurred on Sept 22 2018:
[https://gitlab.com/lv2/lv2/-/commit/4db67120efca2d4c200d2e1ba5cf3d7b97cab97e](url)

Therefore I think we can probably safely do this now.

PR #1262, opened 2024-01-14
Fixed and enhanced Edirol SD-90 IDF

jkl1337:
Current SD-90 IDF does not have LSB set for bank select for any of the patches, so selection among variation does not work.

This version also reorganizes the patch groups so that it is more ergonomic and consistent with other GM2 derived instruments. Patches are grouped by instrument type and then variations from all 4 sets are included in that group sorted by PC and then set (Classic, Contemporary, Solo, Enhanced). This keeps the related variations close together.

The drum sets are added per IDF 2.1 format.
Controller definitions updated.

### DIFF
--- a/src/share/instruments/Edirol-SD90.idf
+++ b/src/share/instruments/Edirol-SD90.idf
@@ -1,1155 +1,907 @@
-<?xml version="1.0"?>
-<muse version="1.0">
-  <MidiInstrument name="Edirol SD-90 StudioCanvas">
-    <PatchGroup name="SD90 Classical Capital Tones">
-      <Patch name="Piano 1" hbank="96" lbank="0" prog="0" />
-      <Patch name="Piano 2" hbank="96" lbank="0" prog="1" />
-      <Patch name="Piano 3" hbank="96" lbank="0" prog="2" />
-      <Patch name="Honky-tonk" hbank="96" lbank="0" prog="3" />
-      <Patch name="E.Piano 1" hbank="96" lbank="0" prog="4" />
-      <Patch name="E.Piano 2" hbank="96" lbank="0" prog="5" />
-      <Patch name="Harpsichord" hbank="96" lbank="0" prog="6" />
-      <Patch name="Clav" hbank="96" lbank="0" prog="7" />
-      <Patch name="Celesta" hbank="96" lbank="0" prog="8" />
-      <Patch name="Glockenspiel" hbank="96" lbank="0" prog="9" />
-      <Patch name="Music Box" hbank="96" lbank="0" prog="10" />
-      <Patch name="Vibraphone" hbank="96" lbank="0" prog="11" />
-      <Patch name="Marimba" hbank="96" lbank="0" prog="12" />
-      <Patch name="Xylophone" hbank="96" lbank="0" prog="13" />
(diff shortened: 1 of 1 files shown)


terminator356 (commented):
Hello. Thank you for the revised IDF file.
However, some changes are required to the file before it can be accepted.

Our IDF files are actually XML files. Certain characters are not allowed directly
 and must be replaced with XML codes.

`&` Must be replaced with `&amp;` 
`<` Must be replaced with `&lt;`
`>` Must be replaced with `&gt;`
`'` Must be replaced with `&apos;`
`"` Must be replaced with `&quot;`

(Note the semicolons at the end of the codes.)

For example:
The original patch `60&apos;s LeadOrg`
 was replaced with `60's LeadOrg`
 but that is actually incorrect and must be changed 
 back to the original `60&apos;s LeadOrg`

jkl1337:
> Hello. Thank you for the revised IDF file. However, some changes are required to the file before it can be accepted.
> 
> Our IDF files are actually XML files. Certain characters are not allowed directly and must be replaced with XML codes.

There are no XML syntax issues with this file.

> `&` Must be replaced with `&amp;` `<` Must be replaced with `&lt;` `>` Must be replaced with `&gt;` `'` Must be replaced with `&apos;` `"` Must be replaced with `&quot;`

This is not correct. See https://www.w3.org/TR/REC-xml/#syntax.  It isn't that they must be used in all contexts. They can.
`>` only needs to be escaped in one context, `&apos;` and `&quot;` are both only necessary in the case of ambiguity in an attribute value.

This file was not produced by hand, it was generated with a python script using it's very mature built-in ElementTree XML implementation and it was both tested with the MusE program and an XML validator (`xmllint`) where there were no issues. If the characters were actually incorrect than one would expect the file not to work in the application.

terminator356 (approved):
I see. Thanks. Approved.
I was under the impression that substitution was required always.

Our XML parser code will accept it as is.
Caveat:
Note however, that our XML saving code is not sophisticated enough
 to deal with contexts that would allow an apostrophe as is.
Thus, our saving code always simply replaces all five of those characters
 with their XML substitutes.
This means if someone opens the instrument in our instrument editor
 and makes changes and then saves a working copy of the instrument,
 those characters will be replaced. [Edit: In the working copy, not the original.]
Should not be a problem... No worries, I suppose...

jkl1337:
> Caveat

I hear you, though one reason I did ignore that is because I took the whitespace lead from the existing `gm2.idf` which puts the drum entries on one line (and I do think that is more readable). When you resave in the application this formatting is busted too.

terminator356:
Likely mostly my fault. Searching "xml substitutions" and then verbosely enforcing them in the code.
Sigh. Rules...
Next time I'm in there I will try to remember to change the saving code.
One hurdle, I call things like `QString::replace('\'', "&apos;");` on every string saved.
Seems I may have to watch for context there and do things different, for ex. if _not_ inside a single quote already,
 if I read those rules correctly.

jkl1337:
Not a criticism, making the in app save look exactly like the curated distribution files doesn't seem to me a pressing issue.
I didn't look at the code, but you're probably OK with that replacement as the character entity replacement is always allowed except in places where it would not be interpreted like a `CDATA` section or the actual markup tokens of course.

terminator356:
Got it. Thanks.

PR #1284, opened 2024-08-30
adding more XG controller definitions

farvardin:
I've added some more definitions for XG instruments, from 

http://studio4all.de/htmle/main91.html
http://www.jososoft.dk/yamaha//pdf/Beggar_SysExGuide.pdf

I've compiled it and it works well (for example for mono / poly on)


terminator356 (approved):
Thanks, approved.

PR #1285, opened 2024-09-26
jack: add timestamps to debug logs

theGreatWhiteShark:
Having timestamps showing at which point in time `Muse4` did encounter which JACK position and triggered which API call is of massive help when correlating the logs with behavior/output of other applications.

The resulting format will look like this

```
[10:20:00.064] timebase_callback is new_pos:1 nframes:512 frame:0 tickPos:0 cpos:0
[10:20:01.062] JackAudioDevice::setMaster val:1 unconditional:0
[10:20:01.067] timebase_callback() state:0
[10:20:01.067] timebase_callback BBT:
 bar:1 beat:1 tick:0
 bar_start_tick:0.000000 beats_per_bar:4.000000 beat_type:4.000000 ticks_per_beat:384.000000 beats_per_minute:120.000000
[10:20:01.067] pos_to_jack_position new: bar:1 beat:1 tick:0
 bar_start_tick:0.000000 beats_per_bar:4.000000 beat_type:4.000000 ticks_per_beat:384.000000 beats_per_minute:120.000000

```

Personally, I am more a fan of wrapping the log code in a function call than a multiline macro, but I did not wanted refactor your code (without asking first).

### DIFF
--- a/src/muse/driver/jack.cpp
+++ b/src/muse/driver/jack.cpp
@@ -35,6 +35,7 @@
 
 #include <QString>
 #include <QStringList>
+#include <QDateTime>
 
 #include <jack/thread.h>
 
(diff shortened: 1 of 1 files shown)


luzpaz:
this is a good idea

PR #1295, opened 2025-02-10
adding definition for casio CDP-S350/S360

farvardin:
I've made a definition for the 700 sounds and drumkits of the Casio CDP-S350 and CDP-S360 pianos. They are listed in the "[Built-in Music Data Lists](https://www.casio.com/content/dam/casio/global/support/manuals/electronic-musical-instruments/pdf/008-en/c/CDPS360_builtinmusicdatalist_EN.pdf)" there: 

https://www.casio.com/intl/electronic-musical-instruments/support.CDP-S360BK/

### DIFF
--- a/src/share/instruments/Casio_CDP-S360.idf
+++ b/src/share/instruments/Casio_CDP-S360.idf
@@ -756,12 +756,6 @@
     <Patch name="700 SFX SET 2" lbank="0" hbank="120" prog="61"  drum="1" />
 </PatchGroup>
 
-  <Drummaps>
-      <entry>
-        <patch_collection lbank="126" prog="1" />
-        <drummap>
 
-
-    </Drummaps>
   </MidiInstrument>
 </muse>
--- a/src/share/instruments/Casio_CDP-S360.idf
+++ b/src/share/instruments/Casio_CDP-S360.idf
@@ -1,0 +1,767 @@
+<?xml version="1.0"?>
+<muse version="2.1">
+  <MidiInstrument name="Casio CDP-S360">
+
+<PatchGroup name="PIANO (CATEGORY : 1)">
+    <Patch name="1 STAGE PIANO" lbank="0" hbank="1" prog="0" />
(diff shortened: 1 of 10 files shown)


src/share/instruments/Casio_CDP-S360.idf
```diff
@@ -0,0 +1,767 @@
...
+    <Patch name="694 LATIN SET 2" lbank="0" hbank="120" prog="50"  drum="1" />
+    <Patch name="695 INDONESIAN SET" lbank="0" hbank="120" prog="53"  drum="1" />
+    <Patch name="696 INDIAN SET" lbank="0" hbank="120" prog="54"  drum="1" />
+    <Patch name="697 ARABIC SET" lbank="0" hbank="120" prog="52"  drum="1" />
+    <Patch name="698 CHINESE SET" lbank="0" hbank="120" prog="51"  drum="1" />
+    <Patch name="699 SFX SET 1" lbank="0" hbank="120" prog="60"  drum="1" />
+    <Patch name="700 SFX SET 2" lbank="0" hbank="120" prog="61"  drum="1" />
+</PatchGroup>
+
+  <Drummaps>
```
terminator356, line 759: Hi. The file seems incomplete. 
There is a drummap 'entry' tag but no closing 'entry' tag.
There is a 'drummap' tag but no closing 'drummap' tag.
The file seems to be missing the actual drum entries.
This file would likely fail to be read by our XML reader.

Is there more to come?
Thanks.

terminator356 (changes requested):
Hi please see the comments in the file.
It seems incomplete and likely would fail to be read by our XML reader.

src/share/instruments/Casio_CDP-S360.idf
farvardin: @terminator356 thank you for the review, there was indeed a xml error, I've removed this part and it is now valid. The drummap seems pretty standard, so I don't think it needs a dedicated section at the moment, it works well this way.

terminator356:
Looks good. Thanks.
If you want to add a drum section later you can do another pull request.

PR #1300, opened 2025-05-24
Update README

farvardin:
extra-cmake-modules is required by ECM (cf. https://linuxmusicians.com/viewtopic.php?t=25418 )


spamatica:
Thanks, added!

PR #1329, opened 2026-06-26
low jitter timer monotonic and various bug fixes

emrum:
### Summary

- This branch also contains the **CLAP host integration.** 

- **Current Status:** works well, more testing recommended. 

- This PR introduces a modern **monotonic high‑resolution timer** with optional real‑time priority, 
replaces legacy POSIX timer usage, 
fixes several stability issues, 
and ensures **correct MIDI PPQN application during file load.**

- **Fix for hexfloat_reading and application of mixer settings.** 

- **Fixes number parsing with different locale setting **

- ** Fix minor compiler warnings **

#### CLAP Plugin Host Integration 
   
-    Synths are supported 
    
-    CLAP plugin GUI - supported. 
    
-    Preset save/restore - supported.

-    Not supported yet: CLAP Effects plugins (muse "plugin" type) . 

#### Modern Monotonic Timer  

-    Replaces older posix_timer logic with a monotonic high‑resolution timer.

-    Uses real‑time priority if permitted by the system for improved jitter and stability.

####  MIDI Division (PPQN) Handling  

-    Correct application of MIDI division (PPQN) during file load to ensure accurate timing resolution.

#### Bug Fixes

-        Fixed CtrlList::read parsing issue

-        Fixed JUCE timer crash in CLAP host integration

-        Fixed parseDouble XML conversion edge cases

-        Additional minor stability fixes
[](url)

### Why This Matters

- These changes improve timing accuracy, reduce jitter, prevent several runtime crashes, and ensure correct MIDI interpretation.


spamatica:
Moving to this pull request instead. I got clap installed and got it past configuration. Infact it compiled too but fails linking, like this:

[100%] Linking CXX executable muse4
/usr/bin/ld: ../libs/plugin/libmuse_plugin_cache_writer_module.so: undefined reference to symbol '_ZN10MusEPlugin21setPluginScanFileInfoERK7QStringPNS_20PluginScanInfoStructE'
/usr/bin/ld: ../libs/plugin/libmuse_plugin_cache_reader_module.so: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [muse/CMakeFiles/muse.dir/build.make:392: muse/muse4] Error 1
make[1]: *** [CMakeFiles/Makefile2:2245: muse/CMakeFiles/muse.dir/all] Error 2
make: *** [Makefile:156: all] Error 2

Any ideas?
It's late I'll try some more tomorrow

emrum:
plugin_cache_reader_module - was not linked in cmake, i have uploaded the fix. 

`git pull   # for update `


note: delete build directory so `cmake -B build` command updates its files.

spamatica:
Alright, I got it to build! Curiously it did not work with the cmake parameters from ./compile_muse.sh, but your suggested command line was fine. I did not investigate but I suspect MODULES_BUILD_STATIC uses different link commands.

Here's some quick feedback before doing more testing:
I continued by downloading uhe Diva as it was the first search result :) 
It appears in the Synth menu in MusE but I could not launch it, the UI freezes.  Now, Diva is a big plugin so it might not be a good first test.

I continued with downloading some random plugins listed on https://clapdb.tech/, namely cchorus.clap and VitaliumVerb.clap. 
Now, both of these should be effect plugins, does your support cover effect plugins yet? They did not appear in the GUI and I can't find any trace that they were scanned.
I would suggest adding printouts the same way that the other plugin scanners do. Basically when the -D switch is used all plugins that are tested are printed (and also print if they failed). 

I added the the plugins to ~/.clap/ and Diva appeared, so that seems like an OK path.

All for the moment, I will test some more later!

emrum:
only synth are supported currently, muse uses a different method for effects type. Diva synth works well here, did you use the Diva/installer.sh ? copying is not enough.

spamatica:
Yes I'm pretty sure I ran the installer. 
In fact, I tested a few others and they all failed... They were detected but either crashed on instantiation or when opening the UI.
I will try more tonight!

spamatica:
Just a quick note. Tried briefly on another computer (same OS) and Diva and few others worked directly!
I don't know what might be different (I did see you had made some changes), will try on the other system later again.

emrum:
Dear @spamatica  , try this branch, it is most advanced now: https://github.com/muse-sequencer/muse/pull/1340

PR #1335, opened 2026-07-16
Clap integration 4

emrum:
this branch supports clap effects in addition to clap synth. 
this branch includes leak fixes. 
this branch includes midi port fixes (alsa and jack ports were randomly cross connected) .


terminator356:
Thanks.
Reviewing. May take some time, but that's OK we have some time before release and then integration,
 to talk about things.

We could move this to the discussions pages later if you were a member, but we'll start here.

Want to mention a couple of important things:

1: 
You might have seen I recently made a QT6 branch. It works, still some minor things to iron out, but it works.
So I usually merge anything new that I add to master into the QT6 branch just to be prepared for errors.
I need to share what my idea for a plan forward was:
I am feeling strong vibrations that after this upcoming MusE-QT5 release, 
 it should be the **last** QT5 release.
After that, all work should really proceed on the QT6 branch.
Except for bug fixes, backports, things we can get away with and so on.
Yes, let's maintain the QT5 branch for now, not abandon it.
At some point, the QT6 branch should become master, and the QT5 code retired to a different branch.
What do you think? @spamatica what do you think?

2:
I've not browsed the whole request but one thing I saw was some work on applying TPQ
 when loading a song. Not sure how far you've gone with that, but I need to say:
For the last month I've been working in that whole TPQ area. 
Here's what I've been up to:
Upon importing a midi file, if the file's TPQ and our TPQ are different, something will have to happen.
One of the two TPQ will have to scale to match the other.
If the file's TPQ is **lower** than ours, we scale the ticks of all the incoming items upward to match our TPQ. 
If the file's TPQ is **higher** than ours, then we have a problem.
We need to raise our TPQ to match the file's higher TPQ.
That means every tick based thing in the song must be scaled upward.
Thus scaling ticks downward is avoided, and no items such as controllers will be 
lost due to identical ticks (which is forbidden).
I plan to offer three choices to the user when the mismatch is detected. 
- Scale upward, whatever it takes, whichever side needs changing (recommended). 
- Allow scaling downward (with warning that some events may be lost or overlap).
- Do nothing.

In addition, this scaling also must occur when the user changes the TPQ in Settings.

I am currently working on the routines to do the scaling. 
It's complicated. Virtually the whole song has to be scaled, and stored in an Undo blob.
Events. Tempo/sig/key lists. Markers. Everything.
With support for undo/redo. Basically required because what if the user made a mistake in scaling? 

And regarding the Settings TPQ: 
You may have noticed there is a global TPQ (the one shown in Settings)
but there is also a song TPQ - _which is actually unused_ !
I plan to relegate the role of global TPQ to just '**Default TPQ**', which will appear in each new song created.
Meanwhile, the song TPQ will be put to work. It will be the TPQ that the whole song uses instead of global TPQ.
This way, if someone changes a song's TPQ, then creates a new song, the TPQ starts with that **global default TPQ**
 and does not start with what the last setting might have been - because the user might not even remember 
 or be aware that they changed it to something previously.

Well, something like that. Hope that made sense.

I am seeing a few oddities in the request. But I'll address them later.

You seem to be a very thorough coder.
For example, I am impressed that you took the time to warn about low memory / locked memory, 
even down to the very dialog that warns about low memory !
C'mon, be honest, did AI help? He he.
At least the comments seem to be all yours !

Cheers, talk later.

emrum:
1. i did not see your qt6 version yet, thanks for pointing it. as this branch includes many improvements, i suggest to rebase on it, if possible. i has barely changes to qt5 setup, thus may remain mostly compatible (?), except a few related leak fixes. 
2. this branch uses TPQ from song (usually saved from global-TPQ) and restores it as global (!) TPQ value, as that is the value used by sequencer. so currently , if song has different TPQ, the global-TPQ gets changed on load. 
3. a TPQ above 192 usually is not required (in 98% of songs there are no notes faster 1/64) , muse already doubles it by default. 
asking the user what to do is a good idea, else i suggest to preserve song-TPQ by default. especially to stay in sync with audio recordings that cannot be rescaled easily (different speed,etc.). while importing different TQP midi , creates rounding errors for the imported midi, but: the offset may remain minimal and easily remain fixable with quantize function, so do not worry to much. 
 4. AI can increase coding speed, but do errors and wrong decisions.  i 've got to remain understanding, careful and test functionality.

emrum:
Tim, i have added your suggestion: there is a default TPQ now in the settings, that gets applied to new songs (beside the active TPQ). The midi import, i did not touch.

spamatica:
> At some point, the QT6 branch should become master, and the QT5 code retired to a different branch.
What do you think? @spamatica what do you think?

I agree, as you said @terminator356, make the coming release the last qt5 and then move master to qt6. 

--
As for this new PR (sorry I get distracted by so many other things I try to do ;) so it takes a bit of time ).

First, I see you made several pull-requests, I think they are building on each other so the latest PR contains all of the changes?
I find this a little concerning as it gets hard to keep track of what I am testing. 

I saw I got a message from you earlier today about the Qt6 PR, which contains this PR but modified for Qt6, I guess? I didn't look at it yet. Mostly because I already started a test with this branch and it's an older machine so i didn't want to bother with qt6.

If I could wish I would want to separate the PRs so they are not dependent on each other. But there are a lot of changes so I understand if that gets hard. Please try to do the description of each PR as detailed as possible at least.

--
Now on to my current test. I have a few different machines to test on though mostly with kubuntu 24.04.
I started this test on an older laptop that still has kubuntu 22.04. 

1. I see the compile_muse.sh has been modernized, which is good. On my machines I cannot build with it though, due to the  -DMODULES_BUILD_STATIC=ON causing the linking to fail (I think I mentioned this earlier). Removing this it builds fine.
2. I can start DIVA Clap now and it works but curiously I can only open the GUI once. If I close it and open it again the window is black.
3. After this I tried again on the machine where the CLAP plugins didn't work (but with different issues) and sad to say it still does not work. The errors looks to be the same as before, some crash or hang on startup and some work a little bit... I had a few clap effects on here so I tried them too. CChorus: The effect works! But I can't open the GUI. VitaliumReverb: crashes muse when I add it.

It is very curious what could make it work so bad on this machine? It is a pretty good machine Ryzen 5, 16gb mem... The last one I tried was a laptop with an i7 but I have also tried another laptop with a ryzen 5 so it is not clear that it is CPU architecture causing issues. I will do more tests.

emrum:
ah , the static build fails ? 
can you send me the error output, else i cannot say much about it. 

the black screen , diva restart, yes, I have had that. should be fixed in newer branches, 
if not let me know. 

try this branch:
https://github.com/muse-sequencer/muse/pull/1340 
branch name is "26-7-23--styling"
(https://github.com/emrum/muse/tree/26-7-23--styling)

emrum:
if you link dynamicly, also make sure to delete old DLL .so files from /usr/local/lib/muse and /usr/local/share/muse . maybe even from /usr/ install directories.

spamatica:
Ah, right. Here is the link error:
[100%] Linking CXX executable muse4
/usr/bin/ld: ../libs/plugin/libmuse_plugin_cache_writer_module.so: undefined reference to symbol '_ZN10MusEPlugin21setPluginScanFileInfoERK7QStringPNS_20PluginScanInfoStructE'
/usr/bin/ld: ../libs/plugin/libmuse_plugin_cache_reader_module.so: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
gmake[2]: *** [muse/CMakeFiles/muse.dir/build.make:393: muse/muse4] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:2249: muse/CMakeFiles/muse.dir/all] Error 2
gmake: *** [Makefile:156: all] Error 2

I'm a bit mystified that you don't get the same.

>  if you link dynamicly, also make sure to delete old DLL .so files from /usr/local/lib/muse and /usr/local/share/muse . maybe even from /usr/ install directories.

Ok, curious. 
Why would they not be replaced? I have never been aware that the .so files were not replaced

spamatica:
Alright, I managed to solve the issue with the computer that could not run clap. It was not a computer that I normally run audio on so it did not have realtime scheduling properly enabled (and it seems several of the plugins require that). 
But there was actually a second error, $LANG was set to a non existent locale, DIVA did NOT like that. A bit funny, nothing else misbehaved to my knowledge.

Anyway, with this issue out of the way I can hopefully do more (real) tests soon!

spamatica:
Just a short update. I have tried a few other plugins both instrument and effect. I want to do more tests but so far it is working very well (apart from the blank screen on UI reopen in this branch) . 
One thing I did notice is that effects seems not to recall their settings? Maybe it's not implemented yet?

PR #1340, opened 2026-07-22
upgrade theme support to use QT6 Qlementine based setup.

emrum:
upgrade theme-support to use Qlementine based setup, 
with additional style support for muse custom widgets. 

due to Qlementine bugs, 
this requires a patched version of Qlementine, that is included .


spamatica:
First time I'm trying to build with Qt6, but I got the qt6 branch compiled.

With this one however I get a lot of errors like:
build_release/synti/vam/ui_vamguibase.h:1028:25: note:   couldn’t deduce template parameter ‘Func2’
 1028 |         QObject::connect(PWS2, &QSlider::valueChanged, LCDNumber1_2_4, &QLCDNumber::display);

Did the needed qt version increase? I have qt6.4.2

emrum:
unfortunately, the style engine requires QT 6.8.0 or higher. (and cmake >= 3.21.0)

emrum:
i will think about changing to another style engine with better support for older versions.

emrum:
also, there is still a uic conversion error left. so updating QT6 may not fully fix this. 
i work on it.

PR #1341, opened 2026-08-26
Roland D110 definition instrument

farvardin:
A new definition for the big brother of the MT-32.
And I've also fixed the "&" symbol for the Casio CDP S360 xml (which made it invalid)


terminator356:
Thanks very much!

PR #1343, opened 2026-09-06
Windows port

luginf:
here is a functional windows port of MusE. Made with LLM (Claude Code). Feel free to use it or check for what it detected for the known problems during the portage. It took many interations with tries and errors for spotting all the problems, but it's working now.

I've managed to run it on windows seven, I haven't tested on other windows but it should work as well.

There is a binary there: https://github.com/luginf/muse/releases/tag/windows-port-v0.1
I've also built a .deb to test if it doesn't break anything on my own Debian system.

**What this adds**: a working native Windows build of MusE (MSYS2/MinGW-w64, Qt5, RtAudio, CMake), including a new native WinMM MIDI backend. User-tested on real hardware (Windows 7, Focusrite Scarlett 18i8, Casio USB-MIDI keyboard): MIDI input/output, recording, playback, and the metronome all confirmed working end-to-end.

**Key fixes**:
- Replaced the GUI↔audio/sequencer thread IPC (anonymous pipes + poll()) with a loopback TCP socket pair (platform_pipe.h) — Windows' QSocketNotifier only supports sockets, not arbitrary pipes, and fcntl(O_NONBLOCK) is a no-op on Windows pipes.
- Fixed a heap corruption crash: SIMD audio buffers allocated with _aligned_malloc() were being released with plain free() instead of _aligned_free().
- Added a native WinMM MIDI backend (driver/winmmmidi.{h,cpp}) since JACK MIDI isn't a viable option for all Windows users — full device I/O, no JACK dependency required.
- Along the way, fixed five separate bugs specific to the new backend: MidiSeq never being constructed on Windows (silently, since ALSA was the only caller of its constructor function), a sysex-buffer re-arm flood, a crash-on-quit race in MIDI device teardown, an invalid raw pipe used by the MIDI thread's internal timer, and — the actual root cause of corrupted recordings — every recorded MIDI event being timestamped 0 instead of a real frame position (matching the approach already used by the ALSA backend).
- Various build-system fixes for MinGW (Windows DLLs not auto-exporting symbols the way ELF .sos do, some third-party libs — LV2, OSC — crashing at startup independent of this work and disabled for now).

**Known limitations**:
- No native VST hosting yet (VST currently only via the VESTIGE Wine-bridge, which doesn't apply on native Windows). Workaround: an external VST host (e.g. SaviHost) + a MIDI loopback driver (e.g. loopMIDI).
- Occasional audio crackling and MIDI-recording slowdowns observed on the (untuned) Windows 7 test machine — not yet root-caused.

Full technical history and diagnosis notes: src/README.win32 on this branch. Experimental build available at the windows-port-v0.1 release tag.

### DIFF
--- a/.github/workflows/windows_build.yml
+++ b/.github/workflows/windows_build.yml
@@ -156,12 +156,18 @@ jobs:
       # real binary explicitly instead of relying on PATH.
       - name: Package runtime deps (windeployqt)
         if: steps.summary.outputs.built == 'true'
+        continue-on-error: true
         run: |
           cd build/muse
           WINDEPLOYQT=$(find /mingw64 -iname "windeployqt*.exe" 2>/dev/null | head -1)
           echo "Using windeployqt at: $WINDEPLOYQT"
           if [ -n "$WINDEPLOYQT" ]; then
-            "$WINDEPLOYQT" muse4.exe
+            # --no-angle/--no-opengl-sw: this mingw64 Qt5 package wasn't
+            # built with ANGLE (GLES-over-D3D) or the software OpenGL
+            # fallback, so windeployqt errors out trying to copy DLLs
+            # that don't exist (libGLESv2.dll) - MusE only needs desktop
+            # OpenGL, which Qt5Gui.dll already covers directly.
+            "$WINDEPLOYQT" --no-angle --no-opengl-sw muse4.exe
           else
             echo "windeployqt not found under /mingw64 - Qt's platform plugin (platforms/qwindows.dll, required to even create a QApplication) won't be packaged. Investigate the mingw-w64-x86_64-qt5-tools package layout if the artifact fails to start standalone."
           fi
--- a/.github/workflows/windows_build.yml
+++ b/.github/workflows/windows_build.yml
@@ -54,7 +54,14 @@ jobs:
             mingw-w64-x86_64-lilv
             mingw-w64-x86_64-rtaudio
             mingw-w64-x86_64-jack2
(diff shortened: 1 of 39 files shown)


terminator356:
Hi, just a quick check in and hello. I've not tried it or examined it yet.
Wow, it's not a fork? It'll still compile on Linux?

I've got a years-old branch on my Windows where I tried to manually make it into a 
 native VS project without Ming.
That got real hard. So I tried Ming. 
But posix-y linux-y stuff got in the way. 
I was disappointed that things got muddy with Ming, different flavours etc. 
I think I concluded that CodeWeavers Crossover would be required for posix support.

Anyway, you can see from the ChangeLog that I have been trying real hard lately to open the
 door ever wider for a Windows build. 
Removing posix-y linux-y stuff.
For example in the 'Spring 2025 large fixes and features' commit.
But I knew there were still a few areas to tackle.
Hope it helped though.

Cheers.

luginf:
@terminator356 yes, it still compiles on linux, I'm using this version on my linux system at the moment, but I haven't tested it exhaustively yet.

I couldn't make it to cross compile on my linux box (it would require to install all the windows dependencies which would be a p.i.t.a.) so I've made it compile on github (using their windows machines).

This code should be able to live by side of the linux version (I hope). I don't have the skills to review this myself (I've just made a quick check but I don't know much of the inner parts of MusE) so if you wish you can check if there is nothing wrong with its integration. You can also keep that into a separate branch to see how it evolves. Maybe it would be safer this way.

